Issue #65, opened 2024-11-02
Hide all progress bars


D4Vinci:
### Describe the feature:

Hi, I appreciate that this library is still not ready for production, but can't we have a verbose mode to hide all progress bars, like these?

<img width="2029" alt="Screenshot 2024-11-02 at 4 16 35 PM" src="https://github.com/user-attachments/assets/52968aa9-a05f-4067-9e71-83d5ec01e959">

I think it's better to have a command line like `camoufox fetch` to download all browserforge files and the rest. I plan to use Camoufox inside a web scraping library, so having progress bars would definitely hurt the user experience for your library and mine.

I think adding all this logic inside `camoufox fetch` would be even better (like `browserforge update` for example)

daijro:
Hello,

In the latest commit, default addons are now downloaded on `camoufox fetch`. I've also removed the unnecessarily verbose progress bars when downloading addons.

D4Vinci:
Nice work dude! Let's close this ticket then.

PR #66, opened 2024-11-02
Adding the option to update browserforge database with `camoufox fetch`

D4Vinci:
This partially solves #65 
Adds the option to update browserforge's database with `camoufox fetch` as below:
```commandline
camoufox fetch --browserforge
```


daijro:
Thanks!

PR #84, opened 2024-11-18
Proxy credentials should be optional.

Nongzhsh:
Proxy credentials should be optional. Fixed https://github.com/daijro/hrequests/issues/72


daijro:
Thanks for the PR.

Issue #125, opened 2024-12-10
Headless option triggreing code relevant for monitors


Pijukatel:
### Describe the bug:

I tried camoufox with headless=True and it raised exception in code related to monitors. That code should not have been executed at all.

Probably due to this condition:
https://github.com/daijro/camoufox/commit/ad556225504e6dfb40fde31af76d1a24941c2f73#diff-3be9f2ffc5cc50fdea8387999673bce3a0153dd3204a6245465e10b12ed764bdR147

### Example exception:

relevant code snippet:

```
from camoufox import AsyncNewBrowser
...
browser=await AsyncNewBrowser(self._playwright, headless=True)
...
```

```
2024-12-10T07:27:07.142Z         File "/usr/local/lib/python3.12/site-packages/camoufox/utils.py", line 501, in launch_options
2024-12-10T07:27:07.143Z           fingerprint = generate_fingerprint(
2024-12-10T07:27:07.145Z                         ^^^^^^^^^^^^^^^^^^^^^
2024-12-10T07:27:07.147Z         File "/usr/local/lib/python3.12/site-packages/camoufox/fingerprints.py", line 135, in generate_fingerprint
2024-12-10T07:27:07.148Z           return FP_GENERATOR.generate(**config)
2024-12-10T07:27:07.148Z                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-12-10T07:27:07.150Z         File "/usr/local/lib/python3.12/site-packages/browserforge/fingerprints/generator.py", line 170, in generate
2024-12-10T07:27:07.150Z           partial_csp = self.partial_csp(
2024-12-10T07:27:07.151Z                         ^^^^^^^^^^^^^^^^^
2024-12-10T07:27:07.152Z         File "/usr/local/lib/python3.12/site-packages/browserforge/fingerprints/generator.py", line 251, in partial_csp
2024-12-10T07:27:07.153Z           if self._is_screen_within_constraints(screen_string, screen)
2024-12-10T07:27:07.154Z              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-12-10T07:27:07.155Z         File "/usr/local/lib/python3.12/site-packages/browserforge/fingerprints/generator.py", line 278, in _is_screen_within_constraints
2024-12-10T07:27:07.159Z           and screen['height'] >= (screen_options.min_height or 0)
[...]

PR #126, opened 2024-12-10
Fix get_screen_cons handling headless incorrectly.

Pijukatel:
**Closes:** #125


iSuslov:
It actually flips the logic for handling headless mode, which breaks how it’s supposed to work. The framework skips things like monitor detection in headless mode on purpose. 

The error you're hitting is more about how the fingerprint is being generated, probably due to the way Camoufox is being used (try `from camoufox.async_api import AsyncCamoufox`).

PR #140, opened 2024-12-17
Encode Camoufox server config before passing into NodeJS Playwright

vihangatheturtle:
Using `launch_server` caused the following error:

```bash
UnicodeEncodeError: 'charmap' codec can't encode characters in position 2695-2696: character maps to <undefined>
```

Fixed by encoding as base64 before writing stdin and decoding from node.


daijro:
Thanks!

PR #153, opened 2025-01-05
Remove print on start of virtual display

TimurKutsenko:
Fixes printing of debug info on virtual display start. 
```py
if self.debug or True:
```
is always resolved to True and prints debug info


daijro:
Thanks. Forgot to change that.

PR #189, opened 2025-02-08
unexpected keyword argument when using proxy-bypass with geoip=True

gsli97:
passing a proxy to camoufox with the "bypass"-argument while using geoip=True runs into an exception. extending the class attributes fixes the problem on my local machine


daijro:
Thanks for the fix 👍

ZeroHans4:
@gsli97 Hey, do you have a way to contact you directly (email, telegram, linkedin)?

gsli97:
@ZeroHans4 Yeah, same username on telegram

PR #198, opened 2025-02-11
Add version selection and listing for cli commands

### LINKED ISSUE #197: Enable selecting and listing build versions in cli command
### Describe the feature:

According to the [comment](https://github.com/daijro/camoufox/issues/173#issuecomment-2641939185), it appears that issues #173 and #186  are caused by the Firefox 135 build.
However, since camoufox’s fetch command always assumes it is retrieving the latest version, users who wish to use the fetch command are forced to use the build version that has the issue.

To resolve this problem, it seems necessary to implement a feature that allows users to choose a specific version among the supported builds. Additionally, a feature to list the available versions would be beneficial.

cda2:
Fixes #197

I haven't been able to test it thoroughly, so if there are any issues with its behavior, please feel free to adjust it as needed.

### DIFF
--- a/pythonlib/camoufox/__main__.py
+++ b/pythonlib/camoufox/__main__.py
@@ -27,11 +27,11 @@ class CamoufoxUpdate(CamoufoxFetcher):
     Checks & updates Camoufox
     """
 
-    def __init__(self) -> None:
+    def __init__(self, specified_version: Optional[str] = None) -> None:
         """
         Initializes the CamoufoxUpdate class
         """
-        super().__init__()
+        super().__init__(specified_version=specified_version)
         self.current_verstr: Optional[str]
         try:
             self.current_verstr = installed_verstr()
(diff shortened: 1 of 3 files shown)


daijro:
Hello, thanks for the PR. Sorry I didn't have a chance to fully test it! I was hesitant to add it since many compatibility and leak issues can be caused by the python library running on older Camoufox versions, since new parameters are added into the browser and in the Playwright library all the time. It's typically better to always use the latest beta, despite there sometimes being some regressions.

I am considering an alternative, to make a user "Camoufox Manager" UI that allows looking up and fetching multiple Camoufox versions and refactoring the library to handle version compatibility differences better. I'm also thinking of separating the giant fonts folder into it's own repository to pull from, since it typically stays the same between versions and isn't worth redownloading hundreds of megabytes each update. I plan on beginning this all in the upcoming weeks/months once my classes are more out of my way. :+1:

cda2:
Hello, at the time, frequent crashes causing programs to malfunction led me to submit the PR.
I acknowledge that the handling of this PR was not ideal, and I totally agree that developing a manager which properly considers version compatibility, as you mentioned, is a more desirable approach.

If there’s any way I can assist in that regard, I’d be more than happy to help.
Thank you for developing such an excellent library.

PR #203, opened 2025-02-13
Minor dev UI improvements

alternativshik:
Minor dev UI improvements: add patch statuses directly to lists for improved usability


daijro:
Thanks!

PR #214, opened 2025-02-27
Improvements

pauliusbaulius:
- Refactor dockerfile
- Fix multibuild.py. Shutil works for cross-platform moves, os.rename doesn't.
- Update MaskConfig.hpp to "cache" CAMOU_CONFIG_<x> instead of reading->validating->parsing each time GetJson is called. With this update my arm64 compilled binary starts in under 2 seconds instead of default start time of ±8 seconds. This also seems to affect page.goto which speeds up drastically aswell.

Will prob take direction of not supporting multiple CAMOU_CONFIG_ vars and use pattern of doing job, closing browser, starting new browser with new fingerprint.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,32 +1,42 @@
 FROM ubuntu:latest
 
-WORKDIR /app
-
-# Copy the current directory into the container at /app
-COPY . /app
-
-# Install necessary packages
-RUN apt-get update && apt-get install -y \
-    # Mach build tools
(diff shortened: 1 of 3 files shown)


daijro:
Hello, why'd you close the PR? I'd be happy to merge it!

pauliusbaulius:
> Hello, why'd you close the PR? I'd be happy to merge it!

I will open new one, still testing.

PR #216, opened 2025-02-28
Build fixes and performance improvements

pauliusbaulius:
- Add missing dep rsync to Dockerfile, at least for me build failed due to rsync.
- Add -p to Dockerfile mkdir to not break image build if /dist already exists
 - Fix multibuild.py: Shutil works for cross-platform moves, os.rename doesn't.
 - Cached CAMOU_CONFIG in GetJson to significantly boost performance (This pretty much solves all perf issues I have encountered and makes it as fast as default playwright ff)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ RUN apt-get update && apt-get install -y \
     # Python
     python3 python3-dev python3-pip \
     # Camoufox build system tools
-    git p7zip-full golang-go aria2 curl \
+    git p7zip-full golang-go aria2 curl rsync \
     # CA certificates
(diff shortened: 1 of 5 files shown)


pauliusbaulius:
@daijro Please test if it doesn't break your flows, I still have flaky builds on both macos and debian, but will investigate further.

daijro:
Thanks, from the looks of it it shouldn't mess with functionality (besides maybe the try/catch causing some potential linking or crashing issues in certain modules `camoucfg` is included in). I'm building a test build to see if everything working, although I might not be able to get back until morning.

PR #218, opened 2025-02-28
Optimize configuration loading: cache JSON

alternativshik:
- Moved environment variable parsing to a dedicated loadConfig() function, called only once during initialization of the global g_jsonConfig.
- Updated GetJson() to return the cached JSON object, preventing repetitive parsing of environment variables on each access.
- Improved performance and simplified configuration data access.

Replace <stdio.h> with <cstdlib> and <cstdio>

- Updated include directives to use modern C++ headers.
- The new headers place standard library functions in the std namespace, enhancing compatibility and type safety as recommended by modern C++ standards.

### DIFF
--- a/additions/camoucfg/MaskConfig.hpp
+++ b/additions/camoucfg/MaskConfig.hpp
@@ -4,7 +4,6 @@ Written by daijro.
 */
 
 #pragma once
-
 #include "json.hpp"
 #include <memory>
 #include <string>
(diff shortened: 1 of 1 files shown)


alternativshik:
Oh, see the almost same fixes in another PR

Ggeshechka:
очень жду обновления

PR #228, opened 2025-03-06
Get IP fix

techinz:
The problem occurred when:
1. Using IPv6 proxies that couldn't connect to the first URL in the list (https://api.ipify.org), because it doesn't support IPv6 and instead of trying to connect to the next URL, an exception was raised.
2. Using some IPv4 proxies that for some reason couldn't connect to the first URL in the list and raised an exception, which instead of trying the next URL, just raised an exception.

Fixed by removing the exception raising in this case, moving the exception to all other cases and adding the exception output at the end, if failed to get IP address.

Also added PyCharm's ".idea" in .gitignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ pythonlib/*.png
 scripts/*.png
 scripts/test*
 .vscode
+.idea
 /tests/*.disabled
 k8s/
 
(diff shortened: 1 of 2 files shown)


daijro:
Thanks for the PR!

PR #250, opened 2025-03-21
Fix unbound exception ip function

DeviousStoat:
Variables assigned in `as` blocks are garbage collected at the end of the block. Hence even if we define `exception = None` the `exception` variable gets cleaned in the `as exception` block. And on the `raise InvalidIP` line we get an `UnboundLocalError` instead of an `InvalidIP` error with the exception information.

You can easily reproduce with something like:
```py
def invalid() -> None:
    exception = None
    try:
        raise ValueError("test")
    except ValueError as exception:
        pass
    raise RuntimeError(f"Failed to get IP address: {exception}")  # UnboundLocalError


invalid()  # UnboundLocalError
```

### DIFF
--- a/pythonlib/camoufox/ip.py
+++ b/pythonlib/camoufox/ip.py
@@ -101,7 +101,7 @@ def public_ip(proxy: Optional[str] = None) -> str:
         "https://ipecho.net/plain",
     ]
 
-    exception = None
+    end_exception = None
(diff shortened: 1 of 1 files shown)


thitoffanelli (approved):
Good one. One of my projects has been facing this.

Issue #242, opened 2025-03-18
get_screen_cons-headless=True sets headed


NoHitCrit:
### Describe the bug:

When trying to use camoufox on a machine that literally has no display, it breaks because window is not set and get_screen_cons returns sizes outside the minimum range.

I initially discovered it while trying to use hrequests:
https://github.com/daijro/hrequests/issues/92

### Recommended change:

![Image](https://github.com/user-attachments/assets/4ba87f24-1627-4cf6-a6f1-6e8759bcba9e)

### To Reproduce:

```
from camoufox.fingerprints import FP_GENERATOR
from camoufox.utils import get_screen_cons
screen = get_screen_cons(headless=True)
FP_GENERATOR.generate(screen=screen)
```

### Ouput:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In [12], line 1
----> 1 FP_GENERATOR.generate(screen=screen)

File /opt/virtualenvs/lib/python3.10/site-packages/browserforge/fingerprints/generator.py:178, in FingerprintGenerator.generate(self, screen, strict, mock_webrtc, slim, **header_kwargs)
    176 if partial_csp:
    177     header_kwargs['user_agent'] = partial_csp['userAgent']
--> 178 headers = self.header_generator.generate(**header_kwargs)
    179 # Extract generated User-Agent
    180 user_agent = get_user_agent(headers)

File /opt/virtualenvs/lib/python3.10/site-packages/browserforge/headers/generator.py:155, in HeaderGenerator.generate(self, browser, os, device, locale, http_version, user_agent, strict, request_dependent_headers)
    131 """
    132 Generates headers using the default options and their possible overrides.
    133 
   (...)
    142     strict (Optional[bool], optional): If true, throws an error if it cannot generate headers based on the input.
    143 """
    145 options = {
    146     'browsers': tuplify(browser),
    147     'os': tuplify(os),
   (...)
    153     'request_dependent_headers': request_dependent_headers,
    154 }
--> 155 generated: Dict[str, str] = self._get_headers(
    156     **{k: v for k, v in options.items() if v is not None}
    157 )
    158 if (options['http_version'] or self.options['http_version']) == '2':
    159     return pascalize_headers(generated)

File /opt/virtualenvs/lib/python3.10/site-packages/browserforge/headers/generator.py:222, in HeaderGenerator._get_headers(self, request_dependent_headers, user_agent, **options)
    218 relaxation_index = next(
[...]

Issue #141, opened 2024-12-17
Browserforge thrown an error when running basic Camoufox code example


ilyazub:
First of all, thank you for creating such a promising tool!

### Describe the bug:

Browserforge thrown a `No headers based on this input can be generated` error when running the basic code from https://camoufox.com/python/usage/. I haven't found a corresponding issue in the Browserforge repo.

What's the best way to fix that issue?

Code:

```python
from camoufox.sync_api import Camoufox

with Camoufox() as browser:
    page = browser.new_page()
```

Exception:

```
$ poetry run python ./test_camoufox_scraper.py coffee
Traceback (most recent call last):
  File "/home/runner/workspace/./test_camoufox_scraper.py", line 3, in <module>
    with Camoufox() as browser:
  File "/home/runner/workspace/.pythonlibs/lib/python3.11/site-packages/camoufox/sync_api.py", line 29, in __enter__
    self.browser = NewBrowser(self._playwright, **self.launch_options)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/workspace/.pythonlibs/lib/python3.11/site-packages/camoufox/sync_api.py", line 86, in NewBrowser
    from_options = launch_options(headless=headless, debug=debug, **kwargs)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/workspace/.pythonlibs/lib/python3.11/site-packages/camoufox/utils.py", line 501, in launch_options
    fingerprint = generate_fingerprint(
                  ^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/workspace/.pythonlibs/lib/python3.11/site-packages/camoufox/fingerprints.py", line 135, in generate_fingerprint
    return FP_GENERATOR.generate(**config)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/workspace/.pythonlibs/lib/python3.11/site-packages/browserforge/fingerprints/generator.py", line 195, in generate
    headers = self.header_generator.generate(**header_kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/workspace/.pythonlibs/lib/python3.11/site-packages/browserforge/headers/generator.py", line 159, in generate
[...]

PR #315, opened 2025-05-29
fix: Return None for screen size if headless

datawookie:
Hi!

Thanks for the work on Camoufox. Many things that were literally impossible previously are now possible thanks to Camoufox.

One issue that I have been hitting my head against is launching Camoufox headless on an EC2 instance. This issue has previously been raised elsewhere:

- https://github.com/daijro/camoufox/issues/141 and
- https://github.com/daijro/camoufox/issues/242.

The second link proposes a simple solution, which I have tested and found to fix the issue.

This MR applies the proposed change to the `get_screen_cons()` function.

Best regards, Andrew.

### DIFF
--- a/pythonlib/camoufox/utils.py
+++ b/pythonlib/camoufox/utils.py
@@ -157,7 +157,7 @@ def get_screen_cons(headless: Optional[bool] = None) -> Optional[Screen]:
     """
     Determines a sane viewport size for Camoufox if being ran in headful mode.
     """
-    if headless is False:
+    if headless is True:
         return None  # Skip if headless
     try:
         monitors = get_monitors()


Alexei17:
Also fixed my issue, wasn't working at all before. Please merge this

jacobcxdev:
Fixed my issue too. +1

afern247:
Is this really needed? Won't companies know it's an actual automation which then defeats the purpose of the browser? I mean... if I see screen none I automatically assume it's a bot...

JWriter20:
Thanks @datawookie. Landed via #756 (merged as 35b45c1), commit e36e0fe, authorship intact.

One follow-up in 0e1f9a8: the call site read `get_screen_cons(headless or has_display(env))`, so a headful run on a real display was treated as headless and lost its monitor bound, undoing 2266f27 (#499). Fixes #242.

Issue #285, opened 2025-04-28
403 Client Error: rate limit exceeded for url: https://api.github.com/repos/daijro/camoufox/releases


arman-bd:
### Description:

I'm trying to write a test case for a project in GitHub Actions, but in most cases, it gets rate-limited when attempting to fetch [release](https://api.github.com/repos/daijro/camoufox/releases) from its repository because there is no `GITHUB_TOKEN` in the request [pkgman.py](https://github.com/daijro/camoufox/blob/main/pythonlib/camoufox/pkgman.py).

What I'm looking for is a simpler way for users to pass down an ENV (GITHUB_TOKEN) to `pkgman.py`, so it doesn't get rate limited when fetching releases or downloading binaries from the GitHub repo.

### Stack trace:

```bash
tests/integration_test.py:347: Failed
----------------------------- Captured stdout call -----------------------------

Making API call...
Downloading model definition files...
input-network.zip             OK!
browser-helper-file.json      OK!
headers-order.json            OK!
header-network.zip            OK!
fingerprint-network.zip       OK!
API Response Status: 503
API Response Content: {"success": false, "data": null, "error": "Unhandled error: Search failed: 403 Client Error: rate limit exceeded for url: https://api.github.com/repos/daijro/camoufox/releases"}
----------------------------- Captured stderr call -----------------------------
2025-04-16 22:32:45,174 - Scraping with Camoufox
Search failed
Traceback (most recent call last):
  File "/Users/runner/work/scraper/scraper/src/scraper.py", line 46, in search
    async with AsyncCamoufox(
  File "/Users/runner/work/scraper/scraper/.venv/lib/python3.10/site-packages/camoufox/async_api.py", line 31, in __aenter__
    self.browser = await AsyncNewBrowser(_playwright, **self.launch_options)
  File "/Users/runner/work/scraper/scraper/.venv/lib/python3.10/site-packages/camoufox/async_api.py", line 88, in AsyncNewBrowser
    from_options = await asyncio.get_event_loop().run_in_executor(
  File "/Users/runner/.local/share/uv/python/cpython-3.10.17-macos-aarch64-none/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/runner/work/scraper/scraper/.venv/lib/python3.10/site-packages/camoufox/utils.py", line 489, in launch_options
    add_default_addons(addons, exclude_addons)
[...]

PR #351, opened 2025-07-10
Use GITHUB_TOKEN if it exists to download Camoufox

PierreMesure:
Related issue : #285

Just a small PR so Camoufox can be downloaded using a Github token. This makes sure that Github doesn't refuse the connection.


arman-bd:
LGTM ✅

PierreMesure:
Cool to see you back @daijro! Hope you're in a better place now! ♥️

PR #355, opened 2025-07-19
docs: Fix file name of `1-leak-fixes.patch`

regseb:
The file is named [_**1**-leak-fixes.patch_](https://github.com/daijro/camoufox/blob/v135.0.1-beta.24/patches/playwright/1-leak-fixes.patch), but in the [table](https://github.com/daijro/camoufox/blob/v135.0.1-beta.24/patches/playwright/README.md#playwright-patches) it's _**0**-leak-fixes.patch_.


JWriter20:
Your fix landed in `v152.0.4-beta.29` via #707, as commit `692529f` ("docs: Fix file name of `1-leak-fixes.patch`").

Closing as superseded — thanks for catching the docs error.

Issue #253, opened 2025-03-21
How to launch_server with a specific profile?


tttern:
Is it possible to run the Camoufox as server with the profile saved on the disk?
I am interested in running the websocket server with persistent_context and user_data_dir

This code works great

```
from camoufox.sync_api import Camoufox

with Camoufox(
    headless=False, 
    persistent_context=True,
    user_data_dir='my-user-data-dir'
) as browser:
    page = browser.new_page()
    ...
```
But the following code has no effect, user-data-dir remains empty

```
from camoufox.server import launch_server

launch_server(
    headless=False,
    port=1234,
    persistent_context=True,
    user_data_dir='my-user-data-dir'
)
```

ev-turenko:
On your client machine try using

```python
# use stored state
context = browser.new_context(storage_state='storage_state.json')
page = context.new_page()
```

```python
# store state
page.context.storage_state(path='storage_state.json')
```

Don't forget to update the storage state every time you are about to close the browser.

Example of usage:
1) Create a function "capture_login"
2) If you can capture login cookies in non-manual mode and all other data – great. Otherwise, leave some triggers to store your storage_state.json. The simplest trigger is:
```python
# store state (after 100 seconds of waiting, it should be enough to pass the auth by hand)
page.wait_for_timeout(100000)
page.context.storage_state(path='storage_state.json')
```
3) Now in your main test function do smthing like this:
```python
if logged_in:
  scrape_things()
else:
  capture_login()
  scrape_things()
page.context.storage_state(path='storage_state.json')
browser.close()
```
4) Now you should be able to use specific profiles.

Same for configs.

It's just a way to implement the something similar to your request.

Also, launch_options doesn't include an option you're trying to pass, so it won't work. Consider implementing the above.
```python
# camoufox > utils.py > launch_options()
def launch_options(
    *,
    config: Optional[Dict[str, Any]] = None,
    os: Optional[ListOrString] = None,
    block_images: Optional[bool] = None,
    block_webrtc: Optional[bool] = None,
    block_webgl: Optional[bool] = None,
    disable_coop: Optional[bool] = None,
    webgl_config: Optional[Tuple[str, str]] = None,
    geoip: Optional[Union[str, bool]] = None,
    humanize: Optional[Union[bool, float]] = None,
    locale: Optional[Union[str, List[str]]] = None,
    addons: Optional[List[str]] = None,
    fonts: Optional[List[str]] = None,
    custom_fonts_only: Optional[bool] = None,
[...]

Issue #161, opened 2025-01-22
Specifying user_data_dir when launching camoufox as a remote websocket server.


jypelle:
```python
launch_server(
    headless=headless,
    geoip=True,
    proxy=proxy_config,
    block_images=block_images,
    persistent_context=True,
    user_data_dir=args.sessionpath,
)
```

Parameters `persistent_context` and `user_data_dir` do not seem to be taken into account when launching camoufox as a remote websocket server.

TonyChen-SH:
same with you.

JWriter20:
Resolved in `v152.0.4-beta.29` via #707, as commit `c61c55f` ("fix(server): reject unservable persistent-context options").

To be precise about the resolution: `persistent_context` / `user_data_dir` are **not** now supported by `launch_server()` — they are rejected with an explicit error. Playwright's `launchServer` routes through `BrowserType.launch()`, and its `PlaywrightServer` only accepts a `preLaunchedBrowser`, so there is no way to expose a persistent `BrowserContext` over a websocket endpoint.

What is fixed is the silent-ignore behaviour you reported: the options are no longer accepted and quietly dropped in favour of a throwaway profile. `launch_server()` now raises a `ValueError` pointing you at `Camoufox(persistent_context=True, ...)` in-process instead.

Closing — thank you for the report.

PR #398, opened 2025-10-11
Fix launch_server, to be able to use persistent_context and user_data_dir

pierorolando1:
## Issue #253 

When attempting to use `persistent_context` and `user_data_dir` parameters in `launch_server()`:

```python
launch_server(
    # ...
    persistent_context=True,
    user_data_dir='my-user-data-dir'
    # ...
)
```

These parameters are completely ignored - they are not read, parsed, or taken into account by the implementation.

## Solution

According to [playwright-core/src/browserServerImpl.ts](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/browserServerImpl.ts), the underlying implementation supports passing `_userDataDir` for persistent contexts.

This PR adds support for these parameters in `launch_server()` to enable persistent context functionality when launching browser servers.

### DIFF
--- a/pythonlib/camoufox/server.py
+++ b/pythonlib/camoufox/server.py
@@ -19,7 +19,7 @@ def camel_case(snake_str: str) -> str:
     if len(snake_str) < 2:
         return snake_str
     camel_case_str = ''.join(x.capitalize() for x in snake_str.lower().split('_'))
-    return camel_case_str[0].lower() + camel_case_str[1:]
+    return ("_" if snake_str[0] == "_" else "") + camel_case_str[0].lower() + camel_case_str[1:]
 
 
 def to_camel_case_dict(data: Dict[str, Any]) -> Dict[str, Any]:
--- a/pythonlib/camoufox/utils.py
+++ b/pythonlib/camoufox/utils.py
@@ -365,6 +365,8 @@ def launch_options(
     i_know_what_im_doing: Optional[bool] = None,
     debug: Optional[bool] = None,
     virtual_display: Optional[str] = None,
+    persistent_context: Optional[bool] = None,
+    user_data_dir: Optional[Union[str, Path]] = None,
     **launch_options: Dict[str, Any],
 ) -> Dict[str, Any]:
     """
@@ -658,5 +660,6 @@ def launch_options(
         "firefox_user_prefs": firefox_user_prefs,
         "proxy": proxy,
         "headless": headless,
+        "_user_data_dir": str(user_data_dir) if user_data_dir else None,
         **(launch_options if launch_options is not None else {}),
     }


kqvanity:
@pierorolando1 
Have you've tested out this solution to be working?

edit: sorry. that fix works with newer version of playwright

kqvanity:
@pierorolando1 

How would you connect to the ws endpoint? So far it's an unresolved issue with playwright (only Chromium based ones due to reuse of context being only supported over cdp connection)

MichalRogowski:
I've been able to connect like
``` 
   launch_server(
        humanize=True,
        fingerprint=fingerprint,
        geoip=True,
        proxy=proxy_config,
        enable_cache=True,
        persistent_context=True,
        user_data_dir=str(profile_dir),
        port=config['port'],
        ws_path=config['wsPath']
    )


and then in my playwright automation I do 
    
const browser = await firefox.connect({ wsEndpoint });

kqvanity:
Can you access the preexisting browser contexts?

When passing user_data_dir to launch_server, it automatically opens up a new context/window, and I was wondering if I can access/reuse this one (since it's the one with passed/desired context)

MichalRogowski:
The bigger problem for me is that `user_data_dir` in the launch_server approach. But I get consistent results when I run it as `with Camoufox() as browser`

kqvanity:
> I get consistent results when I run it as with Camoufox() as browser

Ya but I'm using the camoufox pypi module only on the server-side.

altrof:
Any updates so far? I also thought why launch_server doesn't provide arguments persistent_context and user_data_dir.

kfirfer:
Does the PR really fixed it and able to use persistent profile ? or its just a stubs ?

altrof:
yes, PR fixed it and able to use persistent_context with user_data_dir. I checked it locally.

kfirfer:
> yes, PR fixed it and able to use persistent_context with user_data_dir. I checked it locally.

Thanks I confirm that it is fixed also for me

JWriter20:
Your work landed in `v152.0.4-beta.29` via #707, as commit `c11b920` ("Fix launch_server, to be able to use persistent_context and user_data_dir"), together with `c61c55f` ("fix(server): reject unservable persistent-context options (#161)") which added explicit rejection of the unservable combinations in `pythonlib/camoufox/server.py`.

JWriter20:
Correction to my comment above — I described this too generously, and you deserve the accurate version.

Your commit `c11b920` is in the `beta.29` history, but it did **not** survive as the final behaviour. It was superseded by `c61c55f` ("fix(server): reject unservable persistent-context options"), which took the opposite approach: `launch_server()` now raises a `ValueError` for `persistent_context` and `user_data_dir` rather than supporting them.

The reasoning given there is that Playwright's `launchServer` routes through `BrowserType.launch()`, and its `PlaywrightServer` only accepts a `preLaunchedBrowser` — so there is no way to expose a persistent `BrowserContext` over a websocket endpoint at all. The fix for #161 became "fail loudly instead of silently ignoring the options" rather than "make them work".

So this PR is closed as superseded in the sense that the underlying issue is addressed, but not by your approach. Sorry for the imprecise first comment — and thank you for the report and the work regardless; it's what surfaced that the options were being silently dropped.

PR #415, opened 2025-11-09
Open newpage in a new tab in an existing window instead of in an new window

abhijithp135:
Usage:

```
import {chromium, firefox} from "playwright"
import { launchOptions } from 'camoufox-js';

(async () => {
	const browser = await firefox.launch({
		...await launchOptions({
			headless: false,
		})
	})
	
	const conext = await browser.newContext()
	const page1 = await conext.newPage()
	await page1.goto('https://example.com');
	const page2 = await conext.newPage()
	await page2.goto('https://example.com');
	
	// Wait for 5 secs
	await new Promise(resolve => setTimeout(resolve, 5000))
	await page1.close()
	await page2.close()
	await context.close()
	await browser.close()
})();

```

### DIFF
--- a/additions/juggler/TargetRegistry.js
+++ b/additions/juggler/TargetRegistry.js
@@ -309,6 +309,45 @@ class TargetRegistry {
 
   async newPage({browserContextId}) {
     const browserContext = this.browserContextForId(browserContextId);
+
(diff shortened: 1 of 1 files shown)


JWriter20:
Closing as this is now the current behavior

PR #477, opened 2026-02-02
feat: Add --executable-path flag to the test command

julien-duponchelle:
This allow to pass to Camoufox library the executable-path from the cli test command allowing to use a camoufox from a different folder during development.

Example:
`python -m camoufox test --executable-path="/tmp/camoufox/camoufox-bin"`


icepaq:
Awesome! It'll be some time until we update the pip package but this will be included for when we do!!

PR #504, opened 2026-02-22
Various Patches

icepaq:


KemalIlkilic:
@icepaq  Do we have any estimation on when these various patches will be finished? 
I’m excited to see Camoufox back.
Linux version we use in docker automatically pulls version 135. Would it make sense to try it with 146 or should we stick with 135 for stability for now?
Thank you guys for your effort. 🙏🏻

icepaq:
@KemalIlkilic I'm merging this PR right now. We have a [second one](https://github.com/CloverLabsAI/camoufox/pull/13) being worked on however which includes the python library to insert the patches. That should be merged sometime next week.

PR #507, opened 2026-03-03
fix: only include proxy key in launch options when proxy is not None

dreamsxin:
When proxy is None, the original code still includes "proxy": None in the returned launch options dictionary. This causes an error downstream when the server expects a proxy object but receives null, e.g.:
"Error launching server: proxy: expected object, got null Failed to launch browser."

The fix conditionally adds the proxy key only when proxy is not None, preventing the null value from being passed to the browser launcher.

### DIFF
--- a/pythonlib/camoufox/utils.py
+++ b/pythonlib/camoufox/utils.py
@@ -651,12 +651,20 @@ def launch_options(
     else:
         executable_path = launch_path()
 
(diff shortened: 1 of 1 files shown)


icepaq:
I just noticed that the change was added by Daijro recently. Thanks for the contribution regardless!

PR #562, opened 2026-04-03
Add media:spoof_codecs config option to prevent codec fingerprinting

### LINKED ISSUE #558: canPlayType() and isTypeSupported() leak system codec libraries
`canPlayType()` and `MediaSource.isTypeSupported()` reveal which system codec libraries are installed. Different environments have different FFmpeg versions or VideoToolbox availability, so these APIs return different results for the same browser profile. CreepJS flags this as differing `media.mimeTypes`.

A Linux server with FFmpeg 6.x spoofing a macOS profile still reports a Linux-server codec matrix. Easy correlation signal.

## Root cause

`MP4Decoder::IsSupportedType()` and `MatroskaDecoder::IsSupportedType()` call `PDMFactory::Supports()` which enumerates actual system decoders. Useful for playback decisions, but leaks the host's codec stack to any page that asks.

## Proposed fix

Add a `media:spoof_codecs` config key. When enabled, short-circuit `IsSupportedType()` to return `true` for any valid codec string without consulting `PDMFactory`. Trade-off: claims support for codecs that may not decode. Fine for fingerprint resistance; opt-in so default behaviour is unchanged.

rubenvereecken:
Closes #558

## Description

`canPlayType()` and `MediaSource.isTypeSupported()` leak which system codec libraries are installed by querying `PDMFactory::Supports()`. Different environments → different codec support → fingerprintable. CreepJS detects this as mismatched `media.mimeTypes`.

### The fix

New `media:spoof_codecs` config key. When enabled, bypasses `PDMFactory::Supports()` in `MP4Decoder::IsSupportedType()` and `MatroskaDecoder::IsSupportedType()` — reports all valid codec strings as supported regardless of system decoders.

```json
{ "media:spoof_codecs": true }
```

Default `false` — standard behaviour unchanged.

### Trade-off

The browser claims codec support even when no actual decoder is present. `canPlayType("video/mp4; codecs=hev1...")` → `"probably"` even without HEVC. Actual playback fails at decode time, not query time. Correct trade-off for anti-fingerprinting; users who need real playback should only enable this with matching decoders installed.

## Type of Change

- [x] New feature

### DIFF
--- a/patches/media-codec-spoofing.patch
+++ b/patches/media-codec-spoofing.patch
@@ -1,0 +1,74 @@
+diff --git a/dom/media/mp4/MP4Decoder.cpp b/dom/media/mp4/MP4Decoder.cpp
+index abcdef0001..abcdef0002 100644
+--- a/dom/media/mp4/MP4Decoder.cpp
++++ b/dom/media/mp4/MP4Decoder.cpp
+@@ -7,6 +7,7 @@
+ #include "MP4Decoder.h"
+
+ #include "H264.h"
++#include "MaskConfig.hpp"
+ #include "VPXDecoder.h"
+ #ifdef MOZ_AV1
+ #  include "AOMDecoder.h"
+@@ -133,6 +134,14 @@ bool MP4Decoder::IsSupportedType(const MediaContainerType& aType,
+     return false;
+   }
+
++  // Camoufox: When media:spoof_codecs is enabled, always report MP4 as
++  // supported. This prevents canPlayType()/isTypeSupported() from leaking
++  // which system codec libraries (FFmpeg, VideoToolbox, GStreamer) are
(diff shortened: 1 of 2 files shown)


JWriter20:
Thanks @rubenvereecken. Landed via #756 (merged as 35b45c1), commit 03ac03c, authorship intact.

Two follow-ups were needed to make it work end to end: `media-codec-spoofing.patch` no longer applied to FF152 (its `#include "MaskConfig.hpp"` hunk targeted an include block FF152 reordered, which aborted the whole patch run) — retargeted in 17fe73a. And `media:spoof_codecs` was never declared in `settings/properties.json`, so `validate_config()` silently dropped it and the documented usage did nothing — declared in 0169975, along with a guard that scans for `MaskConfig::Get*("key")` reads missing from the schema. Fixes #558.

PR #595, opened 2026-04-28
fix(virtdisplay): atomic display claim + isolate per-launch DISPLAY env

JWriter20:
## Summary

Two related bugs broke `Camoufox(headless="virtual")` under parallel use in a single Python process:

1. **TOCTOU race in display claim.** `VirtualDisplay` previously scanned `/tmp/.X*-lock` to pick a "free" display number, then invoked Xvfb on `max(existing) + randint(3, 19)`. Two workers could read the same lock-file set, pick the same hint, and both invoke Xvfb on the same display. The losing Xvfb's `bind()` failed with "Server is already active", that worker silently ended up sharing the winner's Xvfb (or had no display at all).

2. **Shared `os.environ` mutation in the launch path.** `utils.py`'s launch helper aliased `env = os.environ` when no env was passed, then did `env['DISPLAY'] = virtual_display`. With multiple `Camoufox(headless="virtual")` calls running concurrently in one process, the last writer won — Firefox children inherited whatever `DISPLAY` happened to be live at the moment Playwright spawned them, causing some workers' browsers to die at startup pointing at another worker's Xvfb (or no Xvfb).

Both bugs are needed to make parallel launches reliable; either alone leaves a hole.

### virtdisplay.py — atomic claim

Reserve the display number **before** spawning Xvfb, using a camoufox-private claim file that Xvfb itself never sees:

1. Compute a baseline above any existing `.X<N>-lock`.
2. For each candidate, skip if `.X<N>-lock` already exists.
3. `O_CREAT | O_EXCL` on `/tmp/.camoufox-X<N>.claim`. First-to-create wins; concurrent callers get `FileExistsError` and try the next number. Up to 50 attempts before raising `CannotExecuteXvfb`.
4. Write the owning PID into the claim file.
5. Spawn Xvfb on the reserved display.
6. Poll `/tmp/.X11-unix/X<N>` (the socket Xvfb only creates on a successful `bind()`) for up to 10s; raise `CannotExecuteXvfb` on timeout or early Xvfb exit.
7. On `kill()`, terminate Xvfb and unlink the claim file.

The claim file is intentionally separate from Xvfb's `.X<N>-lock` so Xvfb's own lock semantics (stale-PID detection, ownership, X.Org format) are untouched.

### virtdisplay.py — stale-claim recovery

If a camoufox process crashes before `kill()` runs, its claim file would leak and that display number would be permanently blocked. Every `_claim_display()` call sweeps `/tmp/.camoufox-X*.claim` first:

- Read the PID from each claim file. `kill(pid, 0)` raising `ProcessLookupError` → owner is dead → unlink.
- Unreadable / malformed claim files (which can briefly appear during the window between `O_CREAT` and the PID being written) are tolerated for a 5-second grace period before being treated as stale, so a concurrent sweep cannot unlink a live in-progress claim.
- Claim files whose PID is owned by another user (`PermissionError`) are left alone.

### utils.py — copy `os.environ` instead of aliasing

`launch_options(...)` now does `env = dict(environ)` instead of `env = environ`, so per-call `env['DISPLAY'] = virtual_display` writes don't leak across concurrent launches in the same process.

### virtdisplay.py — force Mesa software GLX on Xvfb

While here: set `__GLX_VENDOR_LIBRARY_NAME=mesa` and `LIBGL_ALWAYS_SOFTWARE=1` in Xvfb's spawn env. On hosts with NVIDIA drivers installed, libGLvnd loads NVIDIA's GLX provider, which acquires a global rwlock during GL init that can stall Xvfb startup for several seconds under concurrent GPU contention. Xvfb on `1x1x24` never renders anything GPU-accelerated, so software GLX is strictly an improvement.

## Why not `-displayfd`

An earlier revision used Xvfb's own `-displayfd` (let Xvfb's kernel-side `bind()` race auto-increment past collisions, read the chosen number on a pipe). It worked but had two drawbacks:

1. Every spawn paid Xvfb startup cost even on collision (Xvfb forks, retries internally, then reports).
2. The pipe-read with timeout adds startup-path failure modes (EOF, timeout, unparseable output) that need handling and made hangs harder to diagnose.

The claim-file approach resolves contention before `fork()`, so a colliding worker just retries an `os.open()` instead of waiting on a Xvfb spawn.

## Testing

Verified locally on Linux with Xvfb 21.1.11 on a host with NVIDIA drivers (where the latent libGLvnd issue exists). Test suite covers:

**virtdisplay claim correctness**

- 16 threads claiming displays simultaneously → 16 unique displays, 0 errors
- 16 processes claiming displays simultaneously → 16 unique displays, 0 errors
- 8 real Xvfb instances spawned in parallel (0.89s) → 8 distinct live displays, all responsive to `xdpyinfo`
- Single-display lifecycle → spawn, `xdpyinfo` confirms X.Org responding, `kill()` removes both the X11 socket and the claim file

**Stale-claim recovery (5 sub-cases)**

- Claim file with a dead PID is unlinked by the sweep
- Claim file with a live PID is preserved
- Subprocess claims a display, "crashes" without cleanup; next sweep reclaims the slot
- Malformed claim file is preserved within the 5s grace window, removed after
- Normal claim contains the owning PID and is removed by `kill()`

**End-to-end parallel `Camoufox(headless="virtual")`**

Before the env fix: ~25-30% of workers in N=4 batches failed with `Target page closed` / `BrowserType.launch failed` because their Firefox children inherited another worker's `DISPLAY`.

After the fix:

- 8 rounds × N=4 → 32/32 success
- N=8 → 8/8 success in 7.4s
- N=12 → 12/12 success in 8.5s

### DIFF
--- a/pythonlib/camoufox/utils.py
+++ b/pythonlib/camoufox/utils.py
@@ -539,7 +539,11 @@ def launch_options(
     if i_know_what_im_doing is None:
         i_know_what_im_doing = False
     if env is None:
-        env = cast(Dict[str, Union[str, float, bool]], environ)
+        # Copy os.environ rather than aliasing it. Concurrent calls (each
+        # with its own virtual_display) would otherwise race on the shared
+        # os.environ['DISPLAY'] mutation below and leak each other's DISPLAY
+        # into the wrong Firefox child.
+        env = cast(Dict[str, Union[str, float, bool]], dict(environ))
     if isinstance(executable_path, str):
         # Convert executable path to a Path object
         executable_path = Path(abspath(executable_path))
--- a/pythonlib/camoufox/virtdisplay.py
+++ b/pythonlib/camoufox/virtdisplay.py
@@ -1,8 +1,10 @@
+import glob
 import os
+import os.path
+import random
 import subprocess  # nosec
-from glob import glob
-from multiprocessing import Lock
-from random import randrange
+import tempfile
+import time
 from shutil import which
 from typing import List, Optional
 
@@ -13,20 +15,19 @@ from camoufox.exceptions import (
 )
 from camoufox.pkgman import OS_NAME
 
+# How long to wait for Xvfb to bind its X11 socket before giving up.
+SOCKET_READY_TIMEOUT_S = 10.0
+SOCKET_POLL_INTERVAL_S = 0.05
+
 
 class VirtualDisplay:
-    """
-    A minimal virtual display implementation for Linux.
-    """
+    """A minimal virtual display implementation for Linux."""
 
     def __init__(self, debug: Optional[bool] = False) -> None:
-        """
-        Constructor for the VirtualDisplay class (singleton object).
-        """
         self.debug = debug
         self.proc: Optional[subprocess.Popen] = None
         self._display: Optional[int] = None
-        self._lock = Lock()
+        self._claim_path_used: Optional[str] = None
 
     xvfb_args = (
         # fmt: off
@@ -48,9 +49,6 @@ class VirtualDisplay:
 
     @property
     def xvfb_path(self) -> str:
-        """
-        Get the path to the xvfb executable
-        """
         path = which("Xvfb")
         if not path:
             raise CannotFindXvfb("Please install Xvfb to use headless mode.")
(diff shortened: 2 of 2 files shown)


JWriter20:
Superseded by #597 — opening a clean branch instead of force-pushing again over this one. The new PR contains the same atomic-claim fix using Xvfb's `-displayfd` mechanism, plus a 10s read timeout and tests mirroring the `camoufox-js` suite.

PR #599, opened 2026-04-30
Spoof system-ui font resolution to match target OS

### LINKED ISSUE #598: system-ui CSS keyword resolves to platform font, leaking real OS
## Bug

When spoofing macOS from Linux, the CSS generic keyword `system-ui` resolves to the real GTK system font (`Sans`, `Ubuntu`, etc.) instead of a macOS-plausible font. CreepJS detects this via `headless.systemFonts` — it renders text with `font-family: system-ui` and identifies the measured font as Linux-native.

All other font spoofing works: the font enumeration whitelist, font spacing seeds, per-context font lists. This is the last known font-level leak.

## Root cause

`system-ui` resolves through `ResolveGenericFontNames()` → `GetSystemUIFontFamilies()` in `gfx/thebes/gfxPlatformFontList.cpp`. That function calls `LookAndFeel::GetFont(StyleSystemFont::Menu)`, which on Linux queries the GTK theme and returns the real system font.

The existing font patches don't intercept this path:
- `font-hijacker.patch` sets a whitelist in the constructor — only affects named font lookups through `FindAndAddFamiliesLocked`
- `font-list-spoofing.patch` adds per-context filtering in `FindAndAddFamiliesLocked` — but `system-ui` bypasses it entirely, resolving through the generic family path

The bundled fontconfig (`fonts.conf`) maps `system-ui` → `"System Font"`, but `"System Font"` isn't a real installed font, so fontconfig falls back to `Sans`.

## Proposed fix

Intercept `GetSystemUIFontFamilies()` at the top — before `LookAndFeel::GetFont` — and read `navigator.platform` from MaskConfig to return the appropriate system font:
- `MacIntel` → `Helvetica`
- `Win32` → `Segoe UI`
- Anything else → fall through to real system font

No new config key needed — co-varies with the existing fingerprint automatically. Happy to put up a PR.

rubenvereecken:
Closes #598

## Related Issue

#598 — `system-ui` CSS keyword resolves to platform font, leaking real OS

## Description

On Linux, the CSS generic keyword `system-ui` resolves via `LookAndFeel::GetFont(StyleSystemFont::Menu)` to the real GTK system font (`Sans`, `Ubuntu`, etc.). When spoofing macOS, this is a direct OS leak — CreepJS's `headless.systemFonts` catches it.

The existing font patches don't cover this path. `font-hijacker.patch` and `font-list-spoofing.patch` both operate on `FindAndAddFamiliesLocked`, which handles named font lookups. The `system-ui` generic keyword resolves through `ResolveGenericFontNames()` → `GetSystemUIFontFamilies()` — a separate code path that calls `LookAndFeel::GetFont` directly.

### The fix

Added an early return in `GetSystemUIFontFamilies()` that reads `navigator.platform` from MaskConfig (already set by the fingerprint) and returns the appropriate system font:

- `MacIntel` → `Helvetica`
- `Win32` → `Segoe UI`
- Linux / no config → falls through to real `LookAndFeel::GetFont` (no change)

No new config key — the fix co-varies with the existing fingerprint. If `navigator.platform` says macOS, `system-ui` resolves to Helvetica. No Python-side changes needed.

### Why these fonts

- **Helvetica**: macOS's sans-serif fallback. San Francisco (the actual system font) isn't available as a named font. Helvetica is in every macOS font list and what the bundled fontconfig maps `sans-serif` to.
- **Segoe UI**: Windows system font since Vista. Named and addressable.

### Verification

Canvas `measureText` with unquoted `system-ui` (the CSS generic keyword) now produces identical widths to `Helvetica` when spoofing macOS. Previously matched `Sans` (Linux leak).

Note: quoted `"system-ui"` in CSS is a literal font family name lookup — different code path, not affected by this patch. The unquoted form is what websites use and what CreepJS measures.

## Type of Change

- [x] Bug fix


icepaq (approved):

icepaq:
@PopcornDev1 keep this change in mind if you ever improve the font fingerprinting stuff

PR #605, opened 2026-05-03
Hide navigator.webdriver via [Pref] gate in WebIDL  

barbatus:
## Summary
                                                                                                                                                                                                                                                                              
  `patches/playwright/1-leak-fixes.patch` already drops `Navigator::Webdriver()` to `return false` at the C++ level — but the `webdriver` attribute on `Navigator` is still **exposed** by WebIDL. Anti-bot platforms like FingerprintJS, DataDome and PerimeterX check the   *presence* of `navigator.webdriver`
                                                                                                                                                                                                                                                                              
  ## Description                                                                                                                                                                                                                                                              
  
  Two new hunks added to `1-leak-fixes.patch`:                                                                                                                                                                                                                                
                                                                  
  1. **`modules/libpref/init/StaticPrefList.yaml`** — declare a new static pref `dom.webdriver.enabled` (default `false`). Required because WebIDL `[Pref="x.y.z"]` resolves at build time to `mozilla::StaticPrefs::x_y_z`;                                                                                                                                                                        
  
  2. **`dom/webidl/Navigator.webidl`** — add `Pref="dom.webdriver.enabled"` to the existing `[Constant, Cached]` extended-attribute bracket on the `webdriver` attribute. When the pref is `false`, the WebIDL binding code generator omits the attribute from the prototype entirely; `'webdriver' in navigator` returns `false`, matching vanilla Firefox.
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                             ## Type of Change                                               
                                                                                                                                                                                                                                                                              
  - [x] Bug fix                                                   
  - [ ] New feature                       
  - [ ] Documentation update
  - [ ] Other                                                                                                                                                                                                                                                                 
                                              
  ## Testing                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                 
  Quick local repro:                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                              
  ```python                                                                                                                                                                                                                                                                   
  import asyncio                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                              
  async def main():
      async with AsyncCamoufox(headless=True) as browser:                                                                                                                                                                                                                     
          await page.goto("about:blank")  
          present = await page.evaluate("'webdriver' in navigator")
          value   = await page.evaluate("navigator.webdriver")                                                                                                                                                                                                                
          print(f"  'webdriver' in navigator = {present}")
          print(f"  navigator.webdriver       = {value}")                                                                                                                                                                                                                     
                                                                  
  asyncio.run(main())                                                                                                                                                                                                                                                         
  ```                                                             
                                                            
  ## Fingerprint Report                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                              
  This PR specifically targets a fingerprint surface (`navigator.webdriver` exposure). Build-tester run pending — happy to attach if a maintainer requests it. Expected delta on the score: positive (one fewer detectable signal).                                           
                                                                                                                                                                                                                                                                              
  <details>                                                       
                                                                  
  (to be attached after build-tester run; happy to coordinate)                                                                                                                                                                                                                
                                                                  
  </details>
                                                                                                                                                                                                                                                                              
  ## Checklist                                
                                                                                                                                                                                                                                                                              
  - [ ] I have linked a related issue above — none open; happy to file one if preferred
  - [ ] ~~Service tests pass~~ — temporarily out of service per template                                                                                                                                                                                                      
  - [ ] Build test passes — pending; this PR adds a new static pref so the build *must* recompile WebIDLPrefs.cpp; would appreciate a maintainer's confirmation that the score improves

### DIFF
--- a/patches/playwright/1-leak-fixes.patch
+++ b/patches/playwright/1-leak-fixes.patch
@@ -1,9 +1,39 @@
+diff --git a/modules/libpref/init/StaticPrefList.yaml b/modules/libpref/init/StaticPrefList.yaml
+--- a/modules/libpref/init/StaticPrefList.yaml
++++ b/modules/libpref/init/StaticPrefList.yaml
+@@ -5487,6 +5487,13 @@
+ #endif
+   mirror: always
+
++# Hide navigator.webdriver attribute when false (Camoufox: anti-fingerprinting,
++# referenced via [Pref] in dom/webidl/Navigator.webidl).
++- name: dom.webdriver.enabled
++  type: bool
++  value: false
++  mirror: always
++
+ # Is support for the Web Audio API enabled?
+ - name: dom.webaudio.enabled
+   type: bool
+
+diff --git a/dom/webidl/Navigator.webidl b/dom/webidl/Navigator.webidl
+--- a/dom/webidl/Navigator.webidl
++++ b/dom/webidl/Navigator.webidl
+@@ -310,7 +310,7 @@
+
+ // https://w3c.github.io/webdriver/webdriver-spec.html#interface
+ interface mixin NavigatorAutomationInformation {
+-  [Constant, Cached]
++  [Constant, Cached, Pref="dom.webdriver.enabled"]
+   readonly attribute boolean webdriver;
+ };
+
 diff --git a/dom/base/Navigator.cpp b/dom/base/Navigator.cpp
 index 9880047350..318d2edba8 100644
 --- a/dom/base/Navigator.cpp
 +++ b/dom/base/Navigator.cpp
 @@ -2316,28 +2316,7 @@ dom::PrivateAttribution* Navigator::PrivateAttribution() {
- 
+
  /* static */
  bool Navigator::Webdriver() {
 -#ifdef ENABLE_WEBDRIVER


icepaq:
Hi! Thanks for opening the PR but looks like this is already covered :)

image

image

barbatus:
@icepaq thanks for getting back. I meant rather `'webdriver' in navigator` is truthy, I think it is also suspicious since FF does not have `webdriver` at all.

JWriter20:
Thanks for the PR, and for pushing back after @icepaq's first reply — but I'm closing this, and I want to give the actual reason rather than just re-pointing at the existing coverage.

The **value** is already handled: `patches/playwright/1-leak-fixes.patch` rewrites `Navigator::Webdriver()` to `return false`.

Your follow-up argument was about *presence* — that `'webdriver' in navigator` being truthy is itself suspicious because "FF does not have `webdriver` at all". That premise isn't right. `navigator.webdriver` is a standardised property from the WebDriver spec and has shipped in Firefox since **63**. In a stock, human-driven Firefox, `'webdriver' in navigator` is `true` and the value is `false`.

So gating the attribute out of WebIDL would make Camoufox differ from stock Firefox in a way that stock Firefox doesn't — it would *manufacture* a tell rather than remove one. A detector checking `'webdriver' in navigator` and finding it absent has learned something real about the browser.

Genuinely appreciate the report; this one just cuts the other way.

Issue #442, opened 2026-01-05
Deterministic Hardware Fingerprinting (Seed Persistence) for Session Reuse


Admix2077:
## Is your feature request related to a problem? Please describe.

**The Problem:**
Currently, `persistent_context=True` successfully persists storage states (Cookies, LocalStorage), but it does **not** persist the randomization seeds used for hardware fingerprinting.

Every time the browser instance is restarted (even with the same `user_data_dir`), Camoufox regenerates:
1.  **Canvas/WebGL/Audio Noise Seeds:** Causing the "Fingerprint ID" / "VisitorID" to change completely.
2.  **Hardware Concurrency:** The CPU core count reported by `navigator.hardwareConcurrency` drifts randomly (e.g., 16 -> 12 -> 14).
3.  **GPU Renderer Strings:** Often switch between different mocked vendors.

**The Impact:**
For automation workflows requiring long-term session reuse, this behavior is problematic. Anti-fraud systems interpret this as "Session Hijacking" because the user identity (Cookies) remains the same, but the device characteristics (Hardware Fingerprint) abruptly change.

## Describe the solution you'd like

I request a mechanism to make the hardware fingerprint **deterministic** based on the profile data.

**Proposed Logic:**
1.  When a `user_data_dir` is initialized, generate the randomization seeds (Canvas/WebGL noise, Hardware Concurrency value, etc.) once.
2.  Save these seeds/configurations into a file within the `user_data_dir` (e.g., `fingerprint_config.json` or inside `user.js`).
3.  On subsequent launches pointing to this directory, **load these existing seeds** instead of generating new random ones.

This would ensure that a specific profile always "looks" like the exact same device, enabling true session persistence.

## Describe alternatives you've considered

1.  **Commercial Anti-detect Browsers + Playwright/Rebrowser:**
    * *Failure:* We attempted to automate popular commercial fingerprint browsers using Playwright and Rebrowser patches. However, the external CDP connection often triggers **Cloudflare** checks immediately. Camoufox's internal integration proves to be much stealthier against Cloudflare, which is why we prefer sticking with Camoufox if the persistence issue can be solved.
2.  **JS Injection (e.g., `page.add_init_script`):**
    * *Limitation:* This fails to cover **Web Workers**, which are commonly used by advanced fingerprinting libraries to detect inconsistencies.
3.  **Keeping the browser open indefinitely:**
    * *Limitation:* Not scalable for resource-constrained environments.

## Additional context

[...]

Issue #378, opened 2025-09-01
A way to preserve fingerprint much like cookies


kqvanity:
### Describe the feature:

I'm working on a project with multiple accounts. I want to maintain the same session info as much as possible.
So far, I've preserved the cookies in a regular json file using an ID for each account.
I'd love to load up the previous browser fingerprint using the same ID as well. Be it saved in a serialized json format or binary. Just the ability to save it/retrieve it using an ID.

### Is your feature request related to a problem? Please describe.

Yes. It's quite related to `persistent_context`, in that it's quite resource hungry in terms of disk space. the `user_data_dir` took up 60M for a single context.

Correct me if I'm wrong, but I don't think I'd need anything besides cookies & browser fingerprint (aside from IP, but I guess it's fine unless I'm rotating it like crazy per single session)

theguy000:
Sorry but no. the fingerprints should be confidential. It should be unique. EVERY TIME. use playwrite.

kqvanity:
> It should be unique

Exactly yes. across different sessions/browser launches. But I want to get the ability to conditionally -much like cookies- persist it. I don't want to keep rotating radically different device configuration/fingerprints. 

see:
- [related](https://github.com/daijro/camoufox/issues/328)
- [browserforge](https://github.com/daijro/browserforge/issues/30#issuecomment-3255785522)

coryking:
Chiming in to mention I'd love this too.  I want to bundle the fingerprint with the storage state.  In otherwords, "lock it in" for a specific set of cookies.   Like @kqvanity says, rotating through radically different config / fingerprints for a "saved session" really upsets some of these things.

Issue #328, opened 2025-06-11
Inconsistent Fingerprint IDs on Multiple Camoufox Runs with Same Args on creepjs


RenatoDinis:
### Describe the bug:

Multiple runs with the same fingerprint and arguments on Camoufox result in different fingerprint IDs on https://abrahamjuliot.github.io/creepjs/. Causing it to always be the first visit with lower score.

### Screenshots:
First Run of the program FP ID 7e98926c77fdb83071f0fba96c504c240ebf3e9b3f07ebaec438277a4cdf5d6f
<img width="768" alt="Image" src="https://github.com/user-attachments/assets/c440b565-a5f2-47f2-83f6-07d79ac0d4c0" />

Second Run of the program FP ID 1150311cf2667690233e48a1958322e458c1a37bab0c3d69713f8195c11c80ee

<img width="761" alt="Image" src="https://github.com/user-attachments/assets/46f4a023-d36d-48ab-9d49-ea625e0c35bf" />

### To Reproduce:

Code
`from camoufox.sync_api import Camoufox
import pickle

# load args from pickle file
with open("args.pkl", "rb") as f:
    args = pickle.load(f)

with Camoufox(**args) as browser:
    page = browser.new_page()
    page.goto("https://abrahamjuliot.github.io/creepjs/")
` 
For the args I'm passing on Camoufox(**args) could not serialize them to json, but here is a print of them

<img width="1167" alt="Image" src="https://github.com/user-attachments/assets/4c0d2872-cf20-4043-ac9e-9f57cbfe1bcf" />

<img width="1180" alt="Image" src="https://github.com/user-attachments/assets/bdcd6fb6-63e3-49cd-9201-6a04c4d7c00a" />

### Version:
Pip package:    v0.4.11
Camoufox:       v135.0.1-beta.24 (Up to date!)

NCGSolutions:
Are you specifying the fonts:spacing_seed to be the same value each time as well?

RenatoDinis:
Thanks for replying @NCGSolutions .
I wasn't passing anything in the config key.

I tried setting fonts:spacing_seed like you suggested, but I still get different fingerprint IDs.
Is there a full list of seeds (like fonts:spacing_seed) that I need to set to get consistent fingerprints across runs?

```
from camoufox.sync_api import Camoufox
import pickle

with open("args.pkl", "rb") as f:
    args = pickle.load(f)

args["config"] = {
                "fonts:spacing_seed": 100,
            }

[...]

Issue #71, opened 2024-11-06
when persistent_context=True,restarting the browser,the fingerprint is changed


satanmp:
### Describe the bug:

when persistent_context=True,restarting the browser,the window size and fingerprint is changed

### To Reproduce:

                with Camoufox(persistent_context=True,
                              user_data_dir=user_data_path,
                              # geoip=public_ip,
                              headless=False,
                              i_know_what_im_doing=True,
                              # proxy={"server": proxy_ip},
                              # config={
                              #     'webrtc:ipv4': public_ip,
                              #     'geolocation:latitude': latitude,
                              #     'geolocation:longitude': longitude,
                              # },
                              firefox_user_prefs={'media.peerconnection.enabled': False, },
                              block_webrtc=True,
                              humanize=True,
                              slow_mo=30,
                              # block_images=True,
                              ) as browser:
                    page = browser.new_page()
                    page.goto("https://www.google.com/")

### Version:

beta.15

vanyasem:
Related to https://github.com/daijro/camoufox/issues/38

bai755:
I got the same problem. Unable to remember the same environment variables on the next startup even persistent_context=True

SP-l33t:
@bai755 It's actually somewhat solved [here](https://github.com/daijro/camoufox/issues/38#issuecomment-2799929493)

@daijro Can we get a native option to save fingerprint details to profile folder, please?

JWriter20:
@satanmp This is a duplicate of #38 - both are about persisting/reusing the generated fingerprint (and window size) across sessions with persistent_context. Let's keep the discussion there; it has a working `launch_options` + `from_options` workaround. Closing - comment if you disagree and we'll reopen.

PR #606, opened 2026-05-04
feat: add persistent fingerprint seeds

PopcornDev1:
## Summary

  Adds an optional `fingerprint_seed` parameter for Camoufox-generated
  fingerprints so callers can reproduce stable generated identity values across
  launches and browser contexts.

  Related issues/discussions:

  - Addresses #442
  - Addresses #38
  - Addresses #71
  - Addresses #328
  - Addresses #378
  - Related to discussions #263, #373, #151, and #255

  This is intentionally Python/API-level only. It does not include browser engine,
  C++, TLS, networking, or renderer-level fingerprinting changes.

  ## Motivation

  Camoufox is good at generating fresh fingerprints, but persistent profiles need
  a way to keep generated fingerprint values stable across restarts. Several users
  have reported that cookies/storage can persist while generated fingerprint
  surfaces still rotate, causing CreepJS/fingerprint IDs or browser identity
  signals to change between runs.

  `fingerprint_seed` gives callers a stable generation input while preserving the
  current random-by-default behavior when no seed is supplied.

  ## What Changed

  - Added deterministic seed derivation helpers in `pythonlib/camoufox/` & `fingerprint_seed.py`.
  - Added optional `fingerprint_seed` support to:
    - `launch_options()`
    - sync `NewContext()`
    - async `AsyncNewContext()`
    - `generate_context_fingerprint()`
    - BrowserForge generation/conversion
    - WebGL sampling
  - Supports `str`, `int`, and `bytes` seeds.
  - Rejects unsupported seed types such as `bool`.
  - Uses versioned, length-prefixed SHA-256 material to derive independent
  namespace-specific sub-seeds.
  - Keeps unseeded behavior random.
  - Ensures `fingerprint_seed` is consumed by Camoufox and not forwarded to
  Playwright.
  - Avoids unsupported context-level canvas setter calls in generated init
  scripts.
  - Adds a macOS `.app` bundle fallback for locating `properties.json`.

  ## Seeded Surfaces

  Seeded generation now covers Camoufox-owned generated values including:

  - BrowserForge sampling
  - preset selection
  - OS choice where applicable
  - WebGL sampling
  - font and voice subsets
  - screen offset sampling
  - `window.history.length`
  - audio/canvas/font spacing config seeds

  ## Compatibility / Caveats

  - Existing callers are unaffected unless they pass `fingerprint_seed`.
  - `fingerprint_seed` controls generated fingerprint values only.
  - It does not persist cookies, local storage, cache, or browser profile state.
  - Persistent state should still use a stable `user_data_dir` / persistent
  context.
  - Context-level seeds apply only to surfaces exposed by the current per-context
  init-script path.
  - This PR does not change TLS, proxy, network, or renderer-level behavior.

  ## Testing

  Regression tests added for:

  - deterministic seed derivation
  - namespace/type separation
  - invalid seed input handling
  - sync and async context API forwarding
  - deterministic launch config generation
  - deterministic context fingerprint generation
  - different seeds producing different noise values
  - unseeded launch/context paths remaining randomized
  - BrowserForge stability with concurrent seeded and unseeded calls
  - macOS app bundle `properties.json` fallback

  Local checks run:

  ```bash
  PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=pythonlib python3 -m pytest
  --confcutdir=tests/patches tests/patches/fingerprint-seed.py tests/patches/
  context-api-seed.py tests/patches/context-fingerprint-seed.py tests/patches/
  launch-options-seed.py tests/patches/live-fingerprint-seed.py
```

  Result:

  18 passed, 2 skipped

  ## Dogfooding

  This branch is being dogfooded in VulpineOS for long-lived agent identities:

  - stable profile/account identifier -> fingerprint_seed
  - stable browser profile directory -> user_data_dir
  - generated Camoufox fingerprint values stay reproducible across restarts
  - cookies/storage/profile state remain handled separately by the persistent
    profile layer

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,26 @@ async with AsyncCamoufox() as browser:
 
 [[Installation & usage](https://camoufox.com/python/)]
 
+### Persistent Fingerprint Seeds
+
+Pass `fingerprint_seed` when you want the same generated identity to be reused
+across launches or contexts. The seed controls Camoufox-generated fingerprint
+values such as BrowserForge sampling, font and voice subsets, WebGL sampling,
+and audio/canvas/font noise seeds. It does not persist cookies, storage, or
+browser profile state.
+
+```python
+from camoufox.sync_api import Camoufox, NewContext
+
+with Camoufox(fingerprint_seed="account-123") as browser:
+    page = browser.new_page()
+    page.goto("https://example.com")
+
+    context = NewContext(browser, fingerprint_seed="account-123:context-2")
+    page = context.new_page()
+    page.goto("https://example.com")
+```
+
 ### Making Full use of Hardware Spoofing
 
 For stable releases, you should always use the main [`camoufox`](https://pypi.org/project/camoufox/) pip package. However, if you want to make use of per-context fingerprints and hardware spoofing, use the [`cloverlabs-camoufox`](https://pypi.org/project/cloverlabs-camoufox/) package. This package is updated with each releases, whereas the official package is released on delay.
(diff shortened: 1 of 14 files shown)


PopcornDev1:
@icepaq

PR #608, opened 2026-05-05
fix: per-context WebGL overrides when launch parameters are set

PopcornDev1:
## Summary

  Fixes a WebGL spoofing precedence issue where launch-level `webGl:parameters`
  entries for `UNMASKED_VENDOR_WEBGL` and `UNMASKED_RENDERER_WEBGL` could override
  per-context WebGL vendor/renderer values.

  The affected WebGL constants are:

  - `37445` / `UNMASKED_VENDOR_WEBGL`
  - `37446` / `UNMASKED_RENDERER_WEBGL`

  Previously, `ClientWebGLContext::GetParameter()` checked the generic
  `MaskConfig::GLParam()` path before the existing per-context
  `WebGLParamsManager` path. If the launch fingerprint included
  `webGl:parameters["37445"]` or `webGl:parameters["37446"]`, those global values
  won even when a context had its own WebGL vendor/renderer configured.

  This change skips the generic parameter lookup for those two debug-renderer
  constants so the existing order is preserved:

  1. Require `WEBGL_debug_renderer_info` to be enabled.
  2. Use per-context WebGL vendor/renderer when present.
  3. Fall back to global `webGl:vendor` / `webGl:renderer`.

  ## Testing

  - Separate contexts can report separate WebGL vendor/renderer values.
  - Per-context WebGL overrides launch-level `webGl:vendor` / `webGl:renderer`.
  - Per-context WebGL overrides launch-level `webGl:parameters["37445"]` /
  `["37446"]`.
  - `NewContext` presets apply WebGL vendor/renderer values.

  Verified locally against a rebuilt Camoufox binary:

  ```text
  tests/patches/live-webgl-
  context.py::test_context_webgl_vendor_renderer_are_isolated_and_reusable PASSED
  tests/patches/live-webgl-
  context.py::test_context_webgl_overrides_launch_config_webgl PASSED
  tests/patches/live-webgl-
  context.py::test_context_webgl_overrides_launch_parameter_webgl PASSED
  tests/patches/live-webgl-
  context.py::test_new_context_applies_preset_webgl_vendor_renderer PASSED
```
  4 passed

  Also verified manually with CreepJS:

  - CreepJS saw the context-specific WebGL vendor/renderer.
  - CreepJS did not show the conflicting parent/global launch values.
  - Two separate contexts reported separate WebGL GPU identities.
 
+ Validated with [VulpineOS](https://github.com/VulpineOS/VulpineOS) against a rebuilt local Camoufox binary to cover the browser-agent runtime case: multiple isolated contexts in one launched browser process, each with its own WebGL vendor/renderer. CreepJS observed the context-specific GPU values and did not show the conflicting launch-level `webGl:parameters["37445"]` / `["37446"]` values.

### DIFF
--- a/patches/webgl-spoofing.patch
+++ b/patches/webgl-spoofing.patch
@@ -381,14 +381,21 @@ index 11578f425f..6c77acd7f3 100644
    }
 
    return ret;
-@@ -2114,6 +2172,57 @@ void ClientWebGLContext::GetParameter(JSContext* cx, GLenum pname,
+@@ -2114,6 +2172,64 @@ void ClientWebGLContext::GetParameter(JSContext* cx, GLenum pname,
    const auto& state = State();
 
    // -
 +  std::optional<
 +      std::variant<int64_t, bool, double, std::string, std::nullptr_t>>
 +      data;
-+  data = MaskConfig::GLParam(pname, mIsWebGL2);
++  switch (pname) {
++    case dom::WEBGL_debug_renderer_info_Binding::UNMASKED_RENDERER_WEBGL:
++    case dom::WEBGL_debug_renderer_info_Binding::UNMASKED_VENDOR_WEBGL:
++      break;
++    default:
++      data = MaskConfig::GLParam(pname, mIsWebGL2);
++      break;
++  }
 +
 +  if (data.has_value()) {
 +    const auto& value = data.value();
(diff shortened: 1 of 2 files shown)


PopcornDev1:
@icepaq

PR #637, opened 2026-06-14
fix(webrtc): stop real-IP leak under a proxy; fabricate srflx when blocked

JWriter20:
Two coordinated changes close the WebRTC IP leak for proxied profiles:

settings/camoufox.cfg — constrain ICE so the real WAN/LAN IP can't reach
the JS layer behind a proxy. Keep the host candidate (no_host=false) but
obfuscate its LAN IP via mDNS (.local), emit only the default-route
address, force proxy_only_if_behind_proxy / _if_pbmode so Firefox won't
send a UDP STUN request around a TCP proxy, and route DNS through SOCKS.
Restoring the host candidate also fixes a fingerprint tell: suppressing
it left a single-srflx shape no real Firefox produces.

patches/webrtc-ip-spoofing.patch — read the spoof IP from MaskConfig (the
old path read a never-populated WebRTCIPManager, so real IPs leaked into
srflx candidates), and when no srflx forms because STUN is blocked behind
the proxy, fabricate a synthetic srflx carrying the proxy IP into the
candidate stream, the SDP, and getStats so WebRTC stops reporting n/a
without ever sending real UDP off-box.

### DIFF
--- a/patches/webrtc-ip-spoofing.patch
+++ b/patches/webrtc-ip-spoofing.patch
@@ -356,21 +356,37 @@ index 648bcda859..c562b95f87 100644
  
    void ForgetSharedWorker(mozilla::dom::SharedWorker* aSharedWorker);
 diff --git a/dom/media/webrtc/jsapi/PeerConnectionImpl.cpp b/dom/media/webrtc/jsapi/PeerConnectionImpl.cpp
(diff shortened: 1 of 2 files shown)


KittisakPh:
@JWriter20 pls back to solve and merge

JWriter20:

PR #638, opened 2026-06-14
fix(proxy): re-enable launch-arg proxy by disabling https_first

JWriter20:
Firefox 146 enables dom.security.https_first by default, which rewrites
plain http:// requests to https:// before the launch-arg proxy filter
sees them. Simple CONNECT-only HTTP proxies (like the Playwright test
fixture) then break. Stock Playwright Firefox builds disable this pref;
do the same in camoufox.cfg.


JWriter20:

PR #640, opened 2026-06-14
fix(stealth): make MaskConfig the single source for screen/device dims

JWriter20:
On FF150 the CSS media-feature device dimensions (device-width /
device-height) no longer matched the spoofed screen.width/height, because
the screen-spoofing patch resolved them from a different source than the
MaskConfig-driven screen object. matchMedia(device-width) therefore
disagreed with screen.width — an internal inconsistency. Re-anchor the
patch so screen and device dimensions both read from MaskConfig.


JWriter20:

PR #642, opened 2026-06-14
fix(juggler): reload about:blank via browsingContext.reload

JWriter20:
Firefox 146's Browser:Reload command is a no-op on about:blank (no
history entry to reload), so Page.reload() silently did nothing there and
the load event never fired (init scripts never re-ran). Fall back to a
forced browsingContext.reload(LOAD_FLAGS_NONE) when the current URI is
about:blank so the load event fires as expected.


JWriter20:

PR #644, opened 2026-06-14
fix(build): create v150-removed dirs before copy-additions

JWriter20:
Firefox 150 removed some source directories that copy-additions.sh copies
into. Without the target dir present, cp fails and the addition is
silently skipped. mkdir -p the v150-removed dirs before copying so the
additions land.


JWriter20:

PR #645, opened 2026-06-14
fix(juggler): filepicker reliability + skip coalescing for synthetic mouse moves

JWriter20:
Two related juggler/input fixes in the playwright patch and PageHandler:

1. File picker: re-anchor the InitFilePicker hunk to the correct function
   (it had drifted onto InitColorPicker on FF150) and fix the page-edge
   mousemove case — a mousemove at exactly x==width or y==height fired as
   an exit event instead of eMouseMove, so the hit-renderer signal never
   arrived and every later input event hung behind it. Treat exact-edge
   coordinates (x >= width || y >= height) as out-of-viewport.

2. Mouse-move coalescing: skip the coalescing branch for juggler-
   synthesized events (mJugglerEventId != 0). Coalescing defers dispatch
   to the next refresh-driver tick, which under heavy parallel load can be
   seconds late and pins activateAndRun's serialization chain. Synthetic
   events now dispatch immediately; real events keep the optimization.


JWriter20:

PR #646, opened 2026-06-14
fix(stealth): make speech-voice spoofing work, and stop leaking host voices

JWriter20:
Two coupled problems left navigator.speechSynthesis exposing the HOST
machine's voices in every spoofed context:

1. MaskConfig::MVoices() requires each `voices` entry to be a full
   {lang,name,voiceUri,isDefault,isLocalService} object and silently
   dropped anything else. The pythonlib emitted voices as bare
   "Name:lang:type" strings, so NOTHING ever registered — for any OS —
   and Firefox fell back to the host's speech-dispatcher / SAPI / NSSpeech
   voices. Linux was additionally skipped on the false premise that it has
   "no native speech voices" (it runs espeak-ng via speech-dispatcher),
   and voices.json `lin` was empty.

Fixes:
- _generate_random_voice_subset now returns MaskConfig voice OBJECTS for
  every target OS (macOS random subset; Windows + Linux full fixed sets),
  with a default voice matching the spoofed locale prefix.
- Linux voiceUris match Firefox's SpeechDispatcherService.cpp byte-for-
  byte: urn:moz-tts:speechd:<name, spaces+non-ASCII escaped>?<lang>.
- Populate voices.json `lin` with the real espeak-ng base-language set
  (131 voices) enumerated by speech-dispatcher.
- _normalize_preset_voices converts presets' legacy string voices to
  objects; setSpeechVoices() extracts names from the objects.
- MVoices() now warns (printf_stderr, like the other getters) and requires
  is_object() instead of silently dropping malformed entries, so this
  class of bug can't hide again.
- Add tests/test_voices.py.


JWriter20:

PR #647, opened 2026-06-14
fix(stealth): correct BrowserForge headless / impossible-geometry tells

JWriter20:
The pythonlib consumes the same BrowserForge fingerprint pool as the
camoufox-js launcher but lacked four leak corrections the JS wrapper
already applies. Each touches only generated values, never user-supplied
config (snapshotted before fingerprint generation):

- fix_navigator_arch: force navigator.platform/oscpu to the UA's arch when
  BrowserForge ships "Linux armv81" against a "Linux x86_64" UA (~8% of the
  Linux pool) — CreepJS cross-checks these and flags the mismatch.
- fix_screen_no_taskbar: when screen.avail == screen on both axes, subtract
  a typical chrome height (mac 25 / win 40 / lin 27) so CreepJS's noTaskbar
  Like-Headless flag doesn't flip; clamp outer/inner height to the new avail.
- clamp_window_dimensions: enforce inner <= outer <= avail <= screen on both
  axes so e.g. outerWidth > screen.width can't leak as impossible geometry.
- set_media_devices_defaults: expose one mic + one camera so an empty
  enumerateDevices() list isn't a headless tell.

Adds tests/test_fingerprint_fixes.py. (Stacked on the voice-spoofing fix.)

---
**Note:** stacked on #646 (voice spoofing fix) — both share `pythonlib/camoufox/fingerprints.py`. Until #646 merges, this PR's diff against `main` will also show #646's voice changes. Review the fingerprint-correction commit in isolation, or merge #646 first.


JWriter20:

PR #655, opened 2026-07-02
Fix playwright session leak when launch or close raises (fixes #82)

### LINKED ISSUE #82: Error: It looks like you are using Playwright Sync API inside the asyncio loop. Please use the Async API instead.
### Describe the bug:

This is a tricky bug for me. It only happens on PC, not on Mac. 
I'm not using any async code. 
I'm not using it in a notebook.
I am using proxies.

### Screenshots:
None

### To Reproduce:
The code loops through records from a database, opens camoufox.
`with Camoufox(addons=[addon_path], proxy={'server': f'http://{proxy}', 'username': '', 'password': ''}, geoip=True) as browser:`
 grabs data, then starts the loop again.

It will run many times, and then I'll get the following error:
**"Error: It looks like you are using Playwright Sync API inside the asyncio loop. Please use the Async API instead."**
and every loop after gives the same error until I restart the program.

### Version:
Pip package:    v0.3.8
Camoufox:       v130.0.1-beta.13 (Latest supported: v132.0-beta.15)

Windows and Mac should be on the same version. Above is the Mac, I am working on getting the Windows version.

hunterjsb:
## Related Issue

Closes #82

## Description

`Camoufox.__enter__`/`__exit__` (and the async variants) can leak the playwright session they wrap, and in the sync API that leak is fatal to the whole process:

In sync playwright, the event loop stays running for the entire life of a session — only a completed `PlaywrightContextManager.__exit__` clears the thread's running-loop marker. If the session is started but its teardown never runs, every later sync `Camoufox()`/playwright start in that thread fails instantly with:

```
Error: It looks like you are using Playwright Sync API inside the asyncio loop.
Please use the Async API instead.
```

…until the process restarts.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other

## Testing

Repro (`repro.py`) — session 1 hits a launch/close failure, session 2 shows whether the process is poisoned:

```python

# session 1: geoip proxy check fails after playwright has started
try:
    with Camoufox(headless=True, geoip=True, proxy={"server": "http://127.0.0.1:9"}) as b:
        pass
except Exception as e:
    print(f"session 1 raised: {type(e).__name__}")

# session 2: no proxy, plain launch
try:
    with Camoufox(headless=True) as b2:
        print(f"session 2 launched OK (firefox {b2.version})")
except Exception as e:
    print(f"session 2 failed: {type(e).__name__}: {e}")
```

## Fingerprint Report

<details>

N/A — pythonlib exception-handling only; no fingerprint-related code touched.

</details>

## Checklist

- [ ] ~~Service tests pass (for python library changes)~~ noted as temporarily out of service
- [ ] Build test N/A (no patch changes)

### DIFF
--- a/pythonlib/camoufox/async_api.py
+++ b/pythonlib/camoufox/async_api.py
@@ -32,13 +32,24 @@ class AsyncCamoufox(PlaywrightContextManager):
 
     async def __aenter__(self) -> Union[Browser, BrowserContext]:
         _playwright = await super().__aenter__()
-        self.browser = await AsyncNewBrowser(_playwright, **self.launch_options)
+        try:
+            self.browser = await AsyncNewBrowser(_playwright, **self.launch_options)
+        except BaseException as e:
+            # Any launch failure (InvalidProxy, missing browser, bad options, ...)
+            # must tear down the playwright session started above so the driver
+            # process/connection is not leaked.
+            await super().__aexit__(type(e), e, e.__traceback__)
+            raise
         return self.browser
 
     async def __aexit__(self, *args: Any):
-        if self.browser:
-            await self.browser.close()
-        await super().__aexit__(*args)
+        # Run the base teardown even if browser.close() raises (e.g. the browser
+        # process already crashed), so the playwright driver/connection is not leaked.
+        try:
+            if self.browser:
+                await self.browser.close()
+        finally:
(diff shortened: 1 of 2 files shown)


daijro:
Thanks for the PR 👍

PR #662, opened 2026-07-08
ci: fix build.yml on GitHub-hosted runners (source wipe, ccache, OOM)

neuregex:
## Problem

`.github/workflows/build.yml` currently fails on GitHub-hosted `ubuntu-24.04` runners at the **Build** step, so no artifacts are produced. There are three independent issues; fixing all three yields a green `windows/x86_64` build (verified on a fork — the resulting `camoufox.exe` launches on Windows 11 with `navigator.platform=Win32` and no WebRTC leak). The same fixes apply to every matrix target.

### 1. `make setup-minimal` wipes the Firefox source on GitHub runners
`setup-minimal` is documented as docker-only (it deliberately skips the local git repo). On a GitHub runner, `actions/checkout` leaves a `.git` at the repo root. `patch.py` then runs, inside the (non-git) source dir:

```
git clean -fdx && ./mach clobber && git reset --hard unpatched
```

With no nested git repo, git escalates to the checkout's `.git` and `git clean -fdx` deletes the entire (untracked) Firefox source, so `set-target` fails with `./mach: not found` and `cp ... services/settings/dumps/main/search-config.json: No such file or directory`.

**Fix:** use `make setup` (which creates the nested git repo + `unpatched` tag that `patch.py` expects) and set a git identity.

### 2. `configure` fails: `Cannot find ccache`
The mozconfig enables `--with-ccache`, but the current `ubuntu-24.04` image no longer ships `ccache`, so configure aborts.

**Fix:** add `ccache` to the apt install step.

### 3. Build OOM (SIGTERM, exit 143) mid-compile
libxul/Rust can exceed the runner's 16 GB RAM + 16 GB swap and the VM is terminated ~46 min in; disk is plentiful (~115 GB free).

**Fix:** bump swap to 32 GB and cap Rust parallelism with `CARGO_BUILD_JOBS=2` (trades build speed for lower peak memory — tune as preferred).

## Verification
With these three changes, `python3 ./multibuild.py --target windows --arch x86_64` completes and uploads `camoufox-*-win.x86_64.zip`, restoring the automated Windows builds that have been missing from CI.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -82,7 +82,9 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install -y msitools p7zip-full aria2
+          # ccache: the mozconfig enables --with-ccache, but the runner image no
+          # longer ships it, so configure fails with "Cannot find ccache".
+          sudo apt-get install -y msitools p7zip-full aria2 ccache
 
       - name: Fetch source
         env:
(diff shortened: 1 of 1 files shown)


KittisakPh:
Thx u very much 🩷 @icepaq @daijro  pls merge

KittisakPh:
@daijro hi can u merge or test this if it work it will build new FF152 for camoufox sir
recheck for good stealth or not the same pls my god

daijro:
Thank you!

Issue #612, opened 2026-05-13
`Browser.newPage` hangs indefinitely when launched via clean Playwright + Camoufox browser build


4dreadhead:
## Summary

When launching Camoufox through standard Playwright clients (Python `playwright` and Ruby `playwright-ruby-client`), the browser process starts successfully and accepts all setup commands (`Browser.enable`, `createBrowserContext`, `setDefaultViewport` etc.), but the very first `Browser.newPage` call never returns. The client hangs forever waiting for a response.

The browser window itself is alive - the address bar is empty, and on the right side it displays the text `JUGGLER` followed by an id.

## Environment

- OS: Linux (Ubuntu)
- Camoufox: `150.0.2-beta.25`
- Ruby client `playwright-ruby-client`: tried `1.59.x` (latest available, no 1.60 release yet)
- Playwright server (via `npx playwright`): tried `1.59` and `1.60`
- Python client `playwright`: `1.59.0`

## Steps to reproduce

Ruby:

```ruby
Playwright.create(playwright_cli_executable_path: 'npx playwright') do |playwright|
  playwright.firefox.launch(headless: false, executablePath: "path/to/camoufox") do |browser|
    page = browser.new_page   # <-- hangs here forever
    page.goto('https://example.com')
    sleep 1000
  end
end
```

Python:

```python
from playwright.sync_api import sync_playwright
import time

with sync_playwright() as playwright:
    browser = playwright.firefox.launch(
        headless=False,
        executable_path="path/to/camoufox",
    )
    page = browser.new_page()   # <-- hangs here forever
    page.goto("https://example.com")
    time.sleep(1000)
    browser.close()
```

Behavior is identical in both languages and across all Playwright versions tried.

## Expected behavior

`browser.new_page()` returns a page object and navigation proceeds.

## Actual behavior

- Camoufox window opens and stays responsive (can be moved, closed manually).
- Address bar is empty; on the right it shows `JUGGLER <id>`.
- Client hangs indefinitely on `Browser.newPage`.
- No error is reported on the Juggler protocol side — request `id:11` is sent but no response ever arrives.

## Protocol log (sanitized)

Setup completes normally, then `Browser.newPage` is sent and never answered:

```
pw:browser <launching> path/to/camoufox -no-remote -wait-for-browser -foreground \
  -profile /tmp/playwright_firefoxdev_profile-XXXXXX -juggler-pipe -silent
pw:browser <launched> pid=NNNNN
[...]

Issue #279, opened 2025-04-22
Camoufox freezes when asynchronously loading multiple pages


barjin:
### Describe the bug:

Opening multiple pages concurrently makes Camoufox seemingly freeze. Likely happening somewhere in the browser binary, as we're getting the same issue with `camoufox-js` (originally reported by @vastblast over at https://github.com/apify/camoufox-js/issues/19).

Replacing Camoufox with vanilla Playwright fixes this issue (the code runs and finishes).

### To Reproduce:

```python
from camoufox.async_api import AsyncCamoufox
import asyncio

async def open_page(browser, i: int):
    page = await browser.new_page()
    await page.goto('https://example.com')
    title = await page.title()
    print(f"{i} Loaded page with title: {title}")
    await page.close()

async def main():
    async with AsyncCamoufox(headless=True) as browser:
        await asyncio.gather(
            open_page(browser, 1),
            open_page(browser, 2),
            open_page(browser, 3)
        )

asyncio.run(main())
```

This code will only produce:

```text
3 Loaded page with title: Example Domain
```

and will hang indefinitely.

### Version:

```
Pip package:    v0.4.11
Camoufox:       v135.0.1-beta.24 (Up to date!)
```

algorri94:
I've seen this behaviour as well. There must be a deadlock when creating a new page. If you use a semaphore to create the pages sequentially, then it works as expected

hashim-neustreet:
The same issue happens when creating multiple browser contexts i.e `browser.new_context()`, asynchronously

nikosep:
@barjin any update on this ?
Did you solve it somehow ?
I have the same issue.

Deferede:
I encountered the same issue: when creating contexts asynchronously, the browser threads hang.


```from camoufox.async_api import AsyncCamoufox
import asyncio


async def task(name, browser):
    ctx = await browser.new_context()
    page = await ctx.new_page()
    await page.goto("https://example.com")
    print(name, await page.title())
    await page.close()
    await ctx.close()


async def main():
    async with AsyncCamoufox(headless=True) as browser:
        await asyncio.gather(
            task("Worker‑1", browser),
            task("Worker‑2", browser),
            task("Worker‑3", browser),
        )
        await asyncio.gather(
            task("Worker‑1", browser),
            task("Worker‑2", browser),
            task("Worker‑3", browser),
        )
        await asyncio.gather(
            task("Worker‑1", browser),
            task("Worker‑2", browser),
[...]

PR #666, opened 2026-07-12
fix(stealth): consolidated stealth/juggler/build fixes (rebased onto Firefox 152)

JWriter20:
## Overview

This consolidates my previously-separate stealth / juggler / build PRs (**#637–#647**) into a single changeset, **rebased onto the current Firefox 152 base** and validated together via a full local build. I'm closing the individual PRs in favor of this one so they can be reviewed and merged as a coherent, already-integration-tested set.

All changes are confined to `patches/`, `additions/`, and one `scripts/copy-additions.sh` build fix — no changes to the test suite or repo tooling.

## Changes

**WebRTC**
- `fix(webrtc): stop real-IP leak under a proxy; fabricate srflx when blocked` — sanitizes candidate IPs in the SDP, `getStats()`, and default-candidate paths from `MaskConfig`; when ICE produces no public candidate (TCP-proxy case) a synthetic srflx candidate is fabricated so the SDP, `getStats()`, and candidate stream stay consistent (a srflx in the SDP with no matching local-candidate stat is itself a tell). Re-anchored for FF152, which removed the `if (!IsClosed())` wrapper in `GetStats` and added `mRtcpMuxPolicy`.

**Proxy**
- `fix(proxy): re-enable launch-arg proxy by disabling https_first` — `dom.security.https_first` was silently overriding the launch-arg proxy.

**Screen / device dimensions**
- `fix(stealth): make MaskConfig the single source for screen/device dims` — CSS media-feature `device-width`/`device-height` now resolve from the same `MaskConfig` source as `screen.width/height`, so `matchMedia(device-width)` agrees with `screen.width` instead of leaking the real window size.

**Fonts**
- `fix(stealth): spoof CSS2 system-font keyword resolution` (`font-system-fonts-css2.patch`) — resolves CSS2 system-font keywords (`menu`, `caption`, etc.) consistently with the spoofed platform.

**Juggler**
- `fix(juggler): filepicker reliability + skip coalescing for synthetic mouse moves` — synthetic (juggler) mouse events bypass the mouse-move coalescing path so input dispatch can't stall for seconds under heavy parallel load; filepicker interception hardened. Updated for FF152's `RecvRealMouseMoveEvent` signature (`aEvent` → `aMouseEvent`, new `MOZ_ASSERT`s).
- `fix(juggler): emit a single input event for insertText on form fields`.
- `fix(juggler): reload about:blank via browsingContext.reload`.
- `fix(juggler): adapt to FF152 API drift — screenshots, popups, touch, popup-window close` (second commit) — four FF152 platform changes broke driven-browser paths that only surface at runtime:
  - **`Page.screenshot` failed on every call** (`can't access property "document", win is undefined`): FF152 renamed `Node.ownerGlobal` → `documentGlobal` and removed the accessor from `Window` entirely. Fixed the screenshot canvas path (`topChromeWindow` needs no accessor — it *is* the chrome window) plus the other stale call sites (`bringToFront`, both screencast `docShell` lookups, `_linkClicked`).
  - **`page.on('popup')` never fired**: FF152 creates a `window.open()` popup's content-process WindowGlobal actor *before* `TabOpen` registers the PageTarget; the old self-binding actor bailed with no retry, leaving the popup channel permanently unbound. `TargetRegistry` now tracks actors and binds actor↔target from whichever side appears first (mirrors upstream Playwright).
  - **`page.tap()` hung**: FF152 removed `windowUtils.sendTouchEvent` (parallel-array API); touch synthesis now uses `Window.synthesizeTouchEvent`.
  - **Closing a window-features popup crashed `Browser.removeBrowserContext`**: `globalOverlay.js` `closeWindow()` calls `SessionStore.maybeDontRestoreTabs()` unconditionally, which threw on juggler popup windows SessionStore never tracked. New `patches/sessionstore-popup-close-crash.patch` makes it a no-op for untracked windows (matching `onClose`'s guard).

**Locale**
- `fix(locale): propagate launch-arg locale to BrowsingContext`.

**Build**
- `fix(build): create v150-removed dirs before copy-additions` — FF150+ dropped some source dirs `copy-additions.sh` writes into; `mkdir -p` them first so additions land (idempotent, harmless on older bases).

**Fingerprint / voices**
- `fix(stealth): correct BrowserForge headless / impossible-geometry tells` and `fix(stealth): make speech-voice spoofing work, and stop leaking host voices`.

## Validation

Built locally for **Linux x86_64 on Firefox 152.0.4** and tested end-to-end:

- **Patches apply cleanly** — all patches apply to the FF152 source with **zero rejects**; juggler ends up in `browser/omni.ja`.
- **Playwright functionality** — driven via Playwright (build-tester harness) across 8 fingerprint profiles: **1024/1048 checks pass, grade [A]**, with **no WebRTC failures** in any profile (validates the getStats/srflx rebase). Global single-identity profiles: macOS 117/118, Linux 116/117. Hundreds of automated interactions (navigation, clicks, form input) ran without juggler stalls.
- **Fingerprint consistency (sundial / CreepJS)** — **~95% (727–728/763)**; WebRTC, global screen, navigator, WebGL, locale and timezone all report consistent.
- **Upstream Playwright suite (added with the FF152 API-drift commit)** — the full **microsoft/playwright-python v1.60.0** `tests/async/` suite run against the built binary: **1406 passed / 0 failed** (48 skipped; the only deselections are environment-dependent `launch_server` tests that require Playwright's bundled Firefox, which doesn't ship for this host OS). This covers screenshots, popups, tap/touch, input, navigation, frames, and video end-to-end.

A prebuilt Linux binary of this exact changeset (including the FF152 API-drift fixes) is available here: **https://github.com/JWriter20/camoufox/releases/tag/v152.0.4-20260712** (macOS/Windows to follow).

## Known / remaining issues

These are **pre-existing or base-level**, not introduced by this changeset:

1. **Cross-OS font consistency leak** — subtle font markers can still disagree with the claimed OS. Present in upstream as well; tracked as a separate follow-up.
2. **`speechSynthesis` voice-list instability on FF152** — voices are correctly spoofed (no host-voice leak), but repeated `getVoices()` calls return a growing list. The voice patch is unchanged from the v150 line, so FF152 changed `speechSynthesis` internals under it; needs a small follow-up specific to FF152.
3. **Per-context screen dimensions** — this branch intentionally uses a single global `MaskConfig` screen (see the screen fix above), so distinct per-context screen sizes are not applied; global/single-identity screen spoofing is correct.
4. `Document.caretRangeFromPoint` is now present because FF152 legitimately added it (base characteristic, not a leak).

### DIFF
--- a/additions/camoucfg/MaskConfig.hpp
+++ b/additions/camoucfg/MaskConfig.hpp
@@ -295,10 +295,17 @@ MVoices() {
   std::vector<std::tuple<std::string, std::string, std::string, bool, bool>>
       voices;
   for (const auto& voice : data["voices"]) {
-    // Check if voice has all required fields
-    if (!voice.contains("lang") || !voice.contains("name") ||
-        !voice.contains("voiceUri") || !voice.contains("isDefault") ||
-        !voice.contains("isLocalService")) {
+    // Each voice must be a full object with all five fields. A bare string
+    // (e.g. "Name:lang:type") or an object missing a field registers NOTHING
+    // and would silently leave the host's native voices exposed, so warn
+    // loudly instead of dropping it quietly.
+    if (!voice.is_object() || !voice.contains("lang") ||
+        !voice.contains("name") || !voice.contains("voiceUri") ||
+        !voice.contains("isDefault") || !voice.contains("isLocalService")) {
+      printf_stderr(
+          "ERROR: 'voices' entry is not a complete object "
+          "{lang,name,voiceUri,isDefault,isLocalService}; skipping: %s\n",
+          voice.dump().c_str());
       continue;
     }
 
--- a/additions/juggler/JugglerFrameParent.sys.mjs
+++ b/additions/juggler/JugglerFrameParent.sys.mjs
@@ -20,21 +20,17 @@ export class JugglerFrameParent extends JSWindowActorParent {
     if (!this.manager?.isCurrentGlobal)
       return;
 
-    // Only interested in main frames for now.
-    if (this.browsingContext.parent)
-      return;
-
-    this._target = TargetRegistry.instance()?.targetForBrowserId(this.browsingContext.browserId);
-    if (!this._target)
-      return;
-
-    this.actorName = `browser::page[${this._target.id()}]/${this.browsingContext.browserId}/${this.browsingContext.id}/${this._target.nextActorSequenceNumber()}`;
-    this._target.setActor(this);
+    // Firefox 152+: the actor may be created BEFORE the chrome-side PageTarget
+    // exists (notably for window.open popups, where the content WindowGlobal is
+    // created before the `TabOpen` event fires). Delegate to the registry, which
+    // tracks the actor and binds it whenever the target appears — in either
+    // order. The previous implementation looked up the target here and bailed
+    // with no retry when it was missing, leaving the popup's page channel
+    // permanently unbound (Page.ready never sent -> no `popup` event).
+    TargetRegistry.instance()?.onActorCreated(this);
   }
 
   didDestroy() {
-    if (!this._target)
-      return;
-    this._target.removeActor(this);
+    TargetRegistry.instance()?.onActorDestroyed(this);
   }
 }
--- a/additions/juggler/TargetRegistry.js
+++ b/additions/juggler/TargetRegistry.js
@@ -115,6 +115,14 @@ export class TargetRegistry {
     this._userContextIdToBrowserContext = new Map();
     this._browserToTarget = new Map();
     this._browserIdToTarget = new Map();
+    // Firefox 152+: a window.open() popup's content-process WindowGlobal actor
+    // (JugglerFrameParent.actorCreated) can be created BEFORE the chrome-side
+    // `TabOpen` fires and the PageTarget is registered. The old self-binding
+    // actor then found no target and bailed with no retry, so the popup's page
+    // channel never became ready -> `Page.ready` never reached the client ->
+    // `page.on('popup')` never fired. Track actors here so binding happens
+    // whenever either side appears, in either order (matches upstream Playwright).
+    this._browserIdToActor = new Map();
 
     this._proxiesWithClashingAuthCacheKeys = new Set();
     this._browserProxy = null;
@@ -388,6 +396,38 @@ export class TargetRegistry {
   targetForBrowserId(browserId) {
     return this._browserIdToTarget.get(browserId);
   }
+
+  // Called from JugglerFrameParent.actorCreated for every main-frame
+  // WindowGlobalParent. Records the actor and binds it to its PageTarget if the
+  // target already exists; if not, PageTarget's constructor will bind it later.
+  onActorCreated(actor) {
+    // Only interested in main frames for now.
+    if (actor.browsingContext.parent)
+      return;
+    const browserId = actor.browsingContext.browserId;
+    this._browserIdToActor.set(browserId, actor);
+    const target = this._browserIdToTarget.get(browserId);
+    target?.setActor(actor);
+  }
+
+  onActorDestroyed(actor) {
+    // browsingContext can be null once the WindowGlobal is fully torn down.
+    const browserId = actor.browsingContext?.browserId;
+    if (browserId === undefined) {
+      // Fall back to clearing whichever entry points at this actor.
+      for (const [id, a] of this._browserIdToActor) {
+        if (a === actor) {
+          this._browserIdToTarget.get(id)?.removeActor(actor);
+          this._browserIdToActor.delete(id);
+        }
+      }
+      return;
+    }
+    const target = this._browserIdToTarget.get(browserId);
+    target?.removeActor(actor);
+    if (this._browserIdToActor.get(browserId) === actor)
+      this._browserIdToActor.delete(browserId);
+  }
 }
 
 export class PageTarget {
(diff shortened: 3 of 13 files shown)


JWriter20:
@icepaq Merged all of the changes from my previous prs into one for ease. Verified all playwright tests (apart from ones that actively counter the stealth) are passing, and stealth is only improved (there are still some preexisting issues I mentioned above). I started manually reviewing the sections with the cpp patches applied, but honestly I have a lot of other work I need to get done, so am mainly just relying on the fact that all of the playwright tests and my stealth pipeline are passing. I had been using all of these patches in my own production system with the 150 version awhile ago though so I am fairly confident they are correct. I built a linux release here: https://github.com/JWriter20/camoufox/releases, will be building for mac and windows later tonight. 

***Edit, I trusted claude thinking it had ran my pipeline when it said it ran the playwright test suite, turns out it ran just our custom tests, not the full playwright suite. Added full fixes so now it is passing the entire multi-thousand application suite this time (there were issues with screenshots and popups)

KittisakPh:
@JWriter20 
I'm looking forward to the Windows version. Thank you so much for your patience and hard work.

KittisakPh:
@daijro hi we want to merge PR.
plssss check and commit sirr

JWriter20:
> @daijro hi we want to merge PR. plssss check and commit sirr

Running with my build should still work fine for your use-case for now, I appreciate the enthusiasm @KittisakPh, but have some patience. Its a little rude to ping daijro when he no longer wants to actively maintain the project and has had health issues, and I'm sure icepaq has seen the message by now. They're both very busy but I'm sure they'll get to it when they have the time.

KittisakPh:
@JWriter20 ok thx u due

daijro:
I'm okay to merge this. Thanks for the work!

introvert:
Heads-up on a regression that appears to trace back to this PR (the screen/window dimension rework rebased onto FF152).

## Symptom

On **Firefox 152 builds** (reproduced on both `v152.0.4-beta.25` and `v152.0.4-beta.26`), **`Browser.newPage()` hangs forever when the config spoofs BOTH `window.inner*` and `window.outer*`.** The first page opens fine; the *second* page/context creation on the same browser never returns. This matches the reports in #612 and #279.

**Firefox 150 (`v150.0.2-beta.25`) is unaffected** — the exact same config recycles pages in ~150 ms. It is not a Playwright-driver mismatch: it hangs identically with driver **1.52** and **1.60** (and 1.61 won't launch FF152 at all: `Protocol error: Browser.setDefaultViewport`).

## Minimal repro

Any config with all four window keys set, then open a page, close it, and open another:

```
config = {
  "screen.width": 360, "screen.height": 800,
  "window.innerWidth": 360,  "window.innerHeight": 740,
  "window.outerWidth": 360,  "window.outerHeight": 800,   # <-- remove either inner* OR outer* and it works
}
# launch (headless) -> new_page()  (ok)
#                   -> page.close() (ok, fast)
#                   -> new_page()   (HANGS > 45s, and the browser context is then dead:
#                                    subsequent Goto -> "target closed")
```

Dropping **either** the `window.inner*` group **or** the `window.outer*` group makes recycling work again. Spoofing `screen.*` alone is completely fine (and `matchMedia('(device-width: 360px)')` correctly agrees with `screen.width` — the fix in this PR works as intended).

## Root cause (best analysis from the source)

The hang is Juggler's viewport handshake never converging:

- `additions/juggler/TargetRegistry.js` resizes the new window then `await`s confirmation:
  ```js
  this._window.resizeBy(width - innerWidth, height + toolbarTop - innerHeight);
  await this._channel.connect('').send('awaitViewportDimensions', { width, height });
  ```
- `additions/juggler/content/main.js` — `awaitViewportDimensions` resolves **only** when
  `domWindow.innerWidth === width && innerHeight === height` (re-checking on `resize` events).

Camoufox's window hijack in `patches/browser-init.patch` pins the inner size via CSS
(`.browserStack { width: innerWidth !important }`) and, when `window.outer*` is set, sizes the
window with `resizeTo(outer)` while **skipping** the inner-reconciliation, guarded by:

```js
if (innerWidth && innerHeight && !(outerWidth || outerHeight)) { window.resizeBy(...); }
```

So when **both** inner and outer are configured, the content inner width can't converge to the value
Juggler asks for on a fresh window → `awaitViewportDimensions` never resolves → `newPage` hangs.

`browser-init.patch` itself is unchanged between 150 and 152, so the "both set" combo was always
latent. What tipped it into a hard deadlock on 152 looks like this PR's screen/window rework — the
old `ScreenDimensionManager` (per-user-context stored dims) was removed in favor of routing
`nsScreen::GetRect()` straight through `MaskConfig` — combined with the FF152 Juggler changes
(`ownerGlobal`→`documentGlobal`, `window.open` actor timing). On 150 the inner/window mismatch
settled; on 152 it no longer does.

## Impact

For headless scrapers that spoof a full window profile (inner + outer) and rotate pages, every page
recycle wedges the worker until an external watchdog kills it — effectively 0 throughput. The only
workarounds today are (a) drop `window.outer*` (which then leaks the real headless `outerWidth`,
i.e. an "outerWidth > screen.width" tell), (b) run under a virtual display (`headless='virtual'`,
where `resizeTo` actually works so the handshake converges), or (c) stay on 150.

## Possible fix direction

Make the reconciliation `resizeBy` run (or otherwise satisfy `awaitViewportDimensions`) even when
`window.outer*` is set — i.e. don't leave the content inner size unable to reach the requested
viewport when both dims are configured. Happy to test a patch against our repro if useful.

Environment: headless, linux-arm64 fleet + local macOS-arm64 repro; Playwright drivers 1.52 and 1.60.

Are you aware of those regressions and what solution would you propose? We don't want to use Xvfb to spoof the screen size.

PR #667, opened 2026-07-14
Add disableInstantAnimations opt-out flag

### LINKED ISSUE #664: Does not display Microsoft's "Press and Hold Button" CAPTCHA properly
When button is "filled" it does not display the checkmark. Why?

Image

pratyush618:
## Issue

Fixes #664 — Microsoft's "Press and Hold" CAPTCHA does not render the checkmark when the button reaches its filled state.

## Root cause

`patches/no-css-animations.patch` (feat: Instant CSS animations, d32f76f) unconditionally collapses every finite CSS animation to zero duration so Playwright doesn't wait on them. Elements that are only revealed *during* an animation's active phase (like the Press-and-Hold checkmark) are therefore never painted — the animation is over before it starts.

The feature is a Playwright-wait convenience, not a fingerprinting defense, so making it opt-out is safe.

## Fix

Introduce a `disableInstantAnimations` config flag, following the existing `disableTheming` pattern:

- **`patches/no-css-animations.patch`** — the finite-animation collapse now runs only when `disableInstantAnimations` is unset. Infinite animations are unaffected, as before.
- **`settings/properties.json`** — registers the `bool` property so it validates through the Python layer.

Default behavior is unchanged (instant animations stay on), so there is no regression. Users hitting animation-dependent UI opt out:

```python
Camoufox(config={"disableInstantAnimations": True})
```

## Behavior

| `disableInstantAnimations` | Finite animation | Infinite animation |
|---|---|---|
| unset (default) | instant (0 duration) | normal |
| `True` | normal | normal |

## Testing

Patch validated by applying against `dom/animation/AnimationEffect.cpp` and confirming the resulting C++ and control flow. A full browser build was not run locally; please run `build-tester` / `make tests` in CI.

### DIFF
--- a/patches/no-css-animations.patch
+++ b/patches/no-css-animations.patch
@@ -2,30 +2,39 @@ diff --git a/dom/animation/AnimationEffect.cpp b/dom/animation/AnimationEffect.c
 index 6ef8c30d49..07cad770c9 100644
 --- a/dom/animation/AnimationEffect.cpp
 +++ b/dom/animation/AnimationEffect.cpp
-@@ -108,10 +108,18 @@ ComputedTiming AnimationEffect::GetComputedTimingAt(
+@@ -13,6 +13,7 @@
+ #include "mozilla/dom/MutationObservers.h"
+ #include "mozilla/dom/ScrollTimeline.h"  // For PROGRESS_TIMELINE_DURATION_MILLISEC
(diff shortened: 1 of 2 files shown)


KittisakPh:
@daijro this’s one look good or bad

daijro:
Thanks. lgtm 👍

PR #669, opened 2026-07-14
Fix uninitialized DateTimeInfo cache in worker timezone override

### LINKED ISSUE #657: Web Worker getTimezoneOffset() intermittently returns 0/garbage while main thread is correct (v150.0.2 — residual of #546)
> **Disclaimer:** I'm a human developer filing this, not a bot. I used an AI assistant
> (Claude) to help investigate and write this report, but the bug is real: the reproduction
> script below and all the numbers were run by me on my own machine, and I've verified the
> output myself. Happy to provide more data if useful.
>
> I searched existing issues first — this looks related to the worker-timezone work in
> #545 / #546 / #556, but is a distinct, still-present signature (see below).

### Describe the bug:

Camoufox spoofs the timezone correctly in the **main thread and in iframes**, but a
**freshly-spawned Web Worker** intermittently returns a wrong value from
`Date.prototype.getTimezoneOffset()` — either `0` (UTC) or a **garbage non-integer**
(e.g. `-0.0000166`, `-4473.92`, `31543.17`). The main thread is always correct.

Notably, **only `getTimezoneOffset()` is affected**: inside the *same* worker,
`Intl.DateTimeFormat().resolvedOptions().timeZone` still reports the correct zone (verified
across 200 workers — 0 `Intl` mismatches vs. ~14% `getTimezoneOffset()` mismatches). So the
worker is internally inconsistent — `Intl` correct, `getTimezoneOffset()` wrong — which points
at the offset-computation path rather than the timezone database.

This appears to be a residual/related case of the worker-timezone work in #545 / #546 / #556,
but with a different signature: those addressed workers leaking the **real system timezone**
(a steady, consistent wrong value). What I see is an **intermittent, transient** bad read —
roughly **12% of workers** — which looks like a race at worker startup, before the worker's
timezone patch is in place.

**Frequency (measured):** roughly **11.7% – 12.4%** of workers return a wrong offset, across
two independent runs. One run of 250 launches × 5 workers = 1250 worker samples gave
**155/1250 (12.4%)** bad — of those, 44 produced a `0` (UTC) read and 114 produced a
non-integer / out-of-range value (some hit both within the same sampling burst). A separate
clean-install run gave **7/60 (~11.7%)**. Over the 250-launch run, the **main thread, child
iframes, and a document-start init script were correct in 100% of cases (0 bad reads)** — so
the leak is specific to Web Workers.

**Why it matters:** it's a fingerprint-consistency leak. The worker's `getTimezoneOffset()`
disagrees both with the main thread and with the worker's *own* `Intl.DateTimeFormat` timezone
— exactly the kind of internal inconsistency anti-bot / anti-fraud vendors look for.

### Screenshots:

Not applicable — this is programmatic; the console output is included under **To Reproduce**.

### To Reproduce:

Prerequisites: Node.js >= 22.

1. Create a fresh folder and install camoufox-js, then download the Camoufox browser:
   ```bash
   mkdir camoufox-tz-repro && cd camoufox-tz-repro
   npm init -y
   npm install camoufox-js
   npx camoufox-js fetch        # downloads the Camoufox browser binary
   ```
   Note: `npx camoufox-js fetch` downloads the Camoufox browser (~311 MB) plus a GeoIP
   database (~66 MB), so on a clean machine the first run can take a few minutes.

2. Save this as `repro.mjs`:
   ```js
   // Compares getTimezoneOffset() in the main thread vs. inside 20 fresh Web Workers.
   // With geoip enabled the main thread reports the timezone matching the public IP.
   // Workers should report the SAME value, but some return 0 or garbage.
   import { Camoufox } from 'camoufox-js';

   // os just sets the spoofed OS string ('windows' | 'macos' | 'linux') — use your own;
   // it shouldn't affect this bug. geoip:true makes the main thread use the IP's timezone.
   const browser = await Camoufox({ headless: true, geoip: true, os: 'macos' });
   const page = await (await browser.newContext({ viewport: null })).newPage();
   await page.goto('about:blank');

   const r = await page.evaluate(async () => {
     const main = new Date().getTimezoneOffset();
     const code = 'onmessage=()=>{const s=new Set();' +
       'for(let k=0;k<5000;k++)s.add(new Date().getTimezoneOffset());' +
       'postMessage([...s]);}';
     const workers = [];
     for (let n = 0; n < 20; n++) {
       const w = new Worker(URL.createObjectURL(new Blob([code], { type: 'application/javascript' })));
       workers.push(await new Promise((res) => { w.onmessage = (e) => res(e.data); w.postMessage(1); }));
     }
     return { main, workers };
   });

   console.log('main thread getTimezoneOffset():', r.main);
   let bad = 0;
   r.workers.forEach((offsets, i) => {
     if (offsets.some((v) => v !== r.main)) { bad++; console.log(`worker #${i}:`, offsets, '<-- MISMATCH'); }
   });
   console.log(`\n${bad}/20 workers disagreed with the main thread (expected 0).`);
   await browser.close();
   ```

3. Run it:
   ```bash
   node repro.mjs
   ```

4. What to look for: the main thread prints one stable offset (e.g. `-180`). Some workers print
   a different value — `0`, a tiny fraction like `-0.0000166`, or a large number like `31543.17`.
   The bug is intermittent (~12% per worker); with 20 workers per run you'll usually see a few
   mismatches. If a given run shows `0/20`, run it a couple more times.

Example output from one run on my machine:
```
main thread getTimezoneOffset(): -180
worker #1:  [ 31543.1683, -180, 18039.08245 ]        <-- MISMATCH
worker #2:  [ -180, -0.00015 ]                        <-- MISMATCH
worker #4:  [ -180, -0.000016666666666666667 ]        <-- MISMATCH
worker #18: [ -180, -9841.239166666666 ]              <-- MISMATCH
worker #19: [ -180, 0 ]                               <-- MISMATCH

5/20 workers disagreed with the main thread (expected 0).
```

**Actual behavior:** a freshly-created `new Worker()` intermittently returns a `getTimezoneOffset()`
that differs from the main thread — `0`, a fraction like `-0.0000166`, or a large out-of-range
number (see output above) — for roughly 12% of workers.

**Expected behavior:** a Web Worker should report the same spoofed timezone / offset as the
main thread, deterministically, on every launch.

### Version:

- Camoufox browser: **150.0.2** — as installed by `npx camoufox-js fetch` (its `version.json` reports release `alpha.25`; current latest release line is v150.0.2)
- camoufox-js: **0.11.1**
- Playwright (via camoufox-js): **1.61.x** (seen 1.61.0 and 1.61.1)
- Node.js: **>= 22** (tested on v25.2.1)
- Python: **n/a** — this uses `camoufox-js` (the JavaScript client), not the Python package
- OS: **macOS 26.1**, Apple Silicon (**arm64**)
- Launch options: `Camoufox({ headless: true, geoip: true, os: 'macos' })` — geoip on, no proxy. The `os` value only sets the spoofed OS string and shouldn't affect this bug; set it to match your platform.

pratyush618:
## Issue

Fixes #657 — inside a freshly-spawned Web Worker, `Date.prototype.getTimezoneOffset()` intermittently (~12%) returns `0` or a garbage non-integer (e.g. `-0.0000166`, `31543.17`), while the main thread and the worker's own `Intl.DateTimeFormat().resolvedOptions().timeZone` stay correct.

## Root cause

`timezone-spoofing.patch` re-adds per-realm timezone overrides on top of the Playwright base, including a constructor:

```cpp
js::DateTimeInfo::DateTimeInfo(const char* timezone) : DateTimeInfo() {
  timeZoneOverride_ = timezone ? std::string(timezone) : std::string();
}
```

Override `DateTimeInfo` instances are used via the fast paths in `DateTimeInfo::getOffsetMilliseconds` / `getDSTOffsetMilliseconds`, which call `internalGet…OffsetMilliseconds` **directly** when a per-realm `dtInfo` is present — bypassing `acquireLockWithValidTimeZone()`, the only path that runs `updateTimeZone()` → `resetState()`. Delegating to the default ctor only sets `timeZoneStatus_ = NeedsUpdate`; it never resets the `RangeCache` members (`dstRange_`, `localRange_`, `utcRange_`) or seeds `utcToLocalStandardOffsetSeconds_`.

So those caches are read uninitialized. In `getOrComputeValue`, if the garbage `startSeconds/endSeconds` happen to bracket the queried time, it returns the garbage `offsetMilliseconds` directly (skipping the correct ICU computation). The `sanityCheck()` `MOZ_ASSERT`s that would catch this are compiled out in the release build. This is why it is intermittent (~12%, uninitialized-memory dependent) and why `Intl` is unaffected — `Intl` reads `timeZoneOverride_` directly.

The upstream per-realm constructor that was removed by the Playwright refactor did exactly this initialization:

```cpp
DateTimeInfo(RefPtr<TimeZoneString> tz)
    : utcToLocalStandardOffsetSeconds_(SecondsPerDay), timeZoneOverride_(tz) {
  resetState();   // updateTimeZone() is never called for override instances
}
```

## Fix

Restore that initialization in the re-added constructor — seed the offset sentinel and call `resetState()`:

```cpp
utcToLocalStandardOffsetSeconds_ = SecondsPerDay;
resetState();
```

Now every override instance starts with valid, empty range caches, so the first `getTimezoneOffset()` is a cache miss that computes the correct offset via ICU, consistent with `Intl` and the main thread.

## Testing

Verified the patch applies cleanly on top of the Playwright-patched `js/src/vm/DateTime.cpp` (3 hunks, no rejects) and produces the intended constructor. A full browser build was not run locally; the repro in #657 (`getTimezoneOffset()` across many fresh workers) should be exercised in CI.

### DIFF
--- a/patches/timezone-spoofing.patch
+++ b/patches/timezone-spoofing.patch
@@ -314,20 +314,28 @@ index 96ed101921..aa78f0fe26 100644
  
  static bool ComputeLocalTime(time_t local, struct tm* ptm) {
    // Neither localtime_s nor localtime_r are required to act as if tzset has
-@@ -256,6 +257,12 @@ js::DateTimeInfo::DateTimeInfo() {
+@@ -256,6 +257,20 @@ js::DateTimeInfo::DateTimeInfo() {
  
  js::DateTimeInfo::~DateTimeInfo() = default;
  
 +js::DateTimeInfo::DateTimeInfo(const char* timezone) : DateTimeInfo() {
 +#if JS_HAS_INTL_API
 +  timeZoneOverride_ = timezone ? std::string(timezone) : std::string();
++
++  // Override instances bypass updateTimeZone()/resetState() (
++  // getOffsetMilliseconds calls them directly when a per-realm dtInfo is
++  // present), so seed the cache state as the upstream per-realm ctor did.
++  // Without this the RangeCache members are read uninitialized and
++  // getTimezoneOffset() intermittently returns 0 or garbage. (issue #657)
++  utcToLocalStandardOffsetSeconds_ = SecondsPerDay;
++  resetState();
 +#endif
 +}
 +
  int64_t js::DateTimeInfo::toClampedSeconds(int64_t milliseconds) {
    int64_t seconds = milliseconds / msPerSecond;
    int64_t millis = milliseconds % msPerSecond;
-@@ -586,6 +593,18 @@ JS_PUBLIC_API bool JS::SetTimeZoneOverride(const char* timeZoneId) {
+@@ -586,6 +601,18 @@ JS_PUBLIC_API bool JS::SetTimeZoneOverride(const char* timeZoneId) {
    return true;
  }
  


KittisakPh:
@daijro here to my brother 🩷

daijro:
Got it, looks good to me. Thanks

Issue #653, opened 2026-06-29
pip install camoufox is broken as of today (new pypi playwright 1.61 release incompatible with bundled Juggler)


albinstman:
Since playwright is unpinned a fresh pip install camoufox now pulls Playwright 1.61.0 which breaks browser launch:

Protocol error (Browser.setDefaultViewport):
Found property "<root>.viewport.isMobile" - false which is not described in this scheme

Cause: Playwright 1.61 added an isMobile field to the viewport sent in Browser.setDefaultViewport, but the bundled Juggler's pageTypes.Viewport schema only knows
viewportSize/deviceScaleFactor.

Affected: both current builds — v135.0.1-beta.24 (stable) and v150.0.2-alpha.25 (prerelease). Pinning playwright==1.60.0 is a working workaround.

Repro:
from camoufox.sync_api import Camoufox
with Camoufox(headless=True) as b:
  b.new_page()  # raises on playwright 1.61.0, fine on 1.60.0

Possible fixes (in order of needed work):
1. Pin playwright>=...,<1.61 in pythonlib
2. Minimal patch: add isMobile: t.Optional(t.Boolean) to pageTypes.Viewport in additions/juggler/protocol/Protocol.js — schema accepts it and handler already ignores it. Verified working with 1.61 via an omni.ja edit on the v150 build.
3. Full update: regenerate the Juggler patch from Playwright 1.61 (+ Firefox 151) so isMobile is actually emulated.

albinstman:
Let me know if you want me to create a PR. Fix 1 requires pypi update (latest update over a year ago), while fix 2 only requires a repackage of omni.ja and a new release.

PierreMesure:
Thank you for reporting @albinstman!

I just [ran into](https://github.com/civictechsweden/g0vse/actions/runs/28638571879) the issue and [pinning Playwright](https://github.com/civictechsweden/g0vse/commit/2b3f76b109909e7182b7be75339569b8c282002d) [fixed it](https://github.com/civictechsweden/g0vse/actions/runs/28640793740) for now.

daijro:
Hello,

Thank you for letting me know. I've added a quick fix in this repo's python library that constrains PW to <1.61.

JWriter20:
[...]

PR #670, opened 2026-07-15
Fix server launch on Playwright 1.60

### LINKED ISSUE #656: Can't launch remote server
### Describe the bug:

I'm on Windows Server 2022 and I run `python -m camoufox server` and it gives error:
```
C:\Users\Administrator>python -m camoufox server
node:internal/modules/cjs/loader:1503
  throw err;
  ^

Error: Cannot find module 'C:\Python314\Lib\site-packages\playwright\driver\package/lib/browserServerImpl.js'
Require stack:
- C:\Python314\Lib\site-packages\camoufox\launchServer.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1500:15)
    at wrapResolveFilename (node:internal/modules/cjs/loader:1071:27)
    at defaultResolveImplForCJSLoading (node:internal/modules/cjs/loader:1095:10)
    at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1116:12)
    at Module._load (node:internal/modules/cjs/loader:1285:25)
    at wrapModuleLoad (node:internal/modules/cjs/loader:255:19)
    at Module.require (node:internal/modules/cjs/loader:1600:12)
    at require (node:internal/modules/helpers:153:16)
    at Object.<anonymous> (C:\Python314\Lib\site-packages\camoufox\launchServer.js:4:39)
    at Module._compile (node:internal/modules/cjs/loader:1854:14) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ 'C:\\Python314\\Lib\\site-packages\\camoufox\\launchServer.js' ]
}

Node.js v24.17.0
BrokenPipeError: [Errno 32] Broken pipe

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<frozen runpy>", line 203, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Python314\Lib\site-packages\camoufox\__main__.py", line 171, in <module>
    cli()
    ~~~^^
  File "C:\Python314\Lib\site-packages\click\core.py", line 1569, in __call__
    return self.main(*args, **kwargs)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "C:\Python314\Lib\site-packages\click\core.py", line 1490, in main
    rv = self.invoke(ctx)
  File "C:\Python314\Lib\site-packages\click\core.py", line 1970, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "C:\Python314\Lib\site-packages\click\core.py", line 1353, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python314\Lib\site-packages\click\core.py", line 907, in invoke
    return callback(*args, **kwargs)
  File "C:\Python314\Lib\site-packages\camoufox\__main__.py", line 131, in server
    launch_server()
    ~~~~~~~~~~~~~^^
  File "C:\Python314\Lib\site-packages\camoufox\server.py", line 65, in launch_server
    process.stdin.close()
    ~~~~~~~~~~~~~~~~~~~^^
OSError: [Errno 22] Invalid argument
```

### To Reproduce:

Run `python -m camoufox server`

### Version:

Pip package:    v0.4.11

pratyush618:
Fixes #656

## Problem

`python -m camoufox server` fails with `MODULE_NOT_FOUND`:

```
Error: Cannot find module '.../playwright/driver/package/lib/browserServerImpl.js'
Require stack:
- .../camoufox/launchServer.js
```

Playwright 1.60 bundled its internals into `lib/coreBundle.js` and removed the private `lib/browserServerImpl.js` that `launchServer.js` required.

The timing is a side effect of #653's mitigation: `195298d` capped `playwright = "<1.61"`, which resolves fresh installs to exactly 1.60.0 — the one release where that file is gone. 1.58 and 1.59 still ship it. The issue was filed the day after the cap landed.

The second traceback in the report (`OSError: [Errno 22]`) is fallout, not a separate bug. Node died, then `process.stdin.close()` raised on the dead pipe and buried the real cause.

## Fix

**Load the driver's package entrypoint instead of its internals.** The driver shipped with playwright-python is a bundled copy of playwright-core, so `index.js` exposes `firefox.launchServer` as public API. Verified present on 1.50.1, 1.58, 1.59 and 1.60, so this stops tracking a private layout that Playwright is free to move.

The driver path is now passed to `launchServer.js` explicitly rather than inferred from `process.cwd()`, and a driver that fails to load reports one actionable line instead of a raw Node stack.

**Use `communicate()` for the config write.** It ignores `BrokenPipeError` and `EINVAL` on both write and close (CPython's `_stdin_write` handles the Windows case explicitly — bpo-19612, bpo-30418), so a server that exits early surfaces its own error and exit code.

The `<1.61` cap is untouched — #653 is an unrelated Juggler protocol-schema bug.

## Tests

Adds `pythonlib/tests/test_server.py`. No browser download or launch; runs in ~1s.

| Test | On this branch | On `main` |
|---|---|---|
| driver entrypoint exposes `launchServer` | pass | pass (contract guard, not a reproducer) |
| `launchServer.js` resolves driver | pass | **fail** — `MODULE_NOT_FOUND` |
| `launch_server` surfaces child exit | pass | **fail** — `BrokenPipeError: [Errno 32]` |

Both failures on `main` are the exact errors from the report. The first test pins the assumption this fix depends on, so the next reshuffle of Playwright's internals fails in CI rather than in a user's terminal.

## Verification

Reproduced and fixed on the reporter's configuration (playwright 1.60.0, Python 3.14):

```
$ python -m camoufox server
Launching server...
Server launched: 616ms
Websocket endpoint: ws://127.0.0.1:37729/31715eea...
```

Connected a real client to that endpoint, drove a page, and confirmed spoofing is intact — `userAgent` reports Windows on a Linux host, `navigator.webdriver` is `False`.

`ruff check`, `ruff format` and `mypy` are clean on the changed files.

Not verified on Windows directly (fix is platform-independent: a JS require path plus stdlib behavior, no Windows-conditional logic). `build-tester` cannot cover this path — it imports only `camoufox.fingerprints`, and its `requirements.txt` pulls the `cloverlabs-camoufox` fork, which replaces this repo's module. `service-tester` needs real proxies.

### DIFF
--- a/pythonlib/camoufox/launchServer.js
+++ b/pythonlib/camoufox/launchServer.js
@@ -1,7 +1,21 @@
-// Workaround that accesses Playwright's undocumented `launchServer` method in Python
-// Without having to use the Node.js Playwright library.
+// Workaround that accesses Playwright's `launchServer` method in Python
+// Without having to install the Node.js Playwright library.
 
-const { BrowserServerLauncherImpl } = require(`${process.cwd()}/lib/browserServerImpl.js`)
+const path = require('path')
+
+// The driver shipped with playwright-python is a copy of playwright-core, so its
+// entrypoint exposes `launchServer`. Resolve through the entrypoint rather than lib/
+// internals, whose layout is private and changes between releases: 1.60 bundled
+// lib/browserServerImpl.js away, which broke this script.
+const driverPackage = process.argv[2]
+
+let playwright
+try {
+    playwright = require(path.join(driverPackage, 'index.js'))
+} catch (error) {
+    console.error(`Error loading the Playwright driver from ${driverPackage}:`, error.message)
+    process.exit(1)
+}
 
 function collectData() {
     return new Promise((resolve) => {
@@ -22,10 +36,7 @@ collectData().then((options) => {
     console.time('Server launched');
     console.info('Launching server...');
     
-    const server = new BrowserServerLauncherImpl('firefox')
-    
-    // Call Playwright's `launchServer` method
-    server.launchServer(options).then(browserServer => {
+    playwright.firefox.launchServer(options).then(browserServer => {
         console.timeEnd('Server launched');
         console.log('Websocket endpoint:\x1b[93m', browserServer.wsEndpoint(), '\x1b[0m');
         // Continue forever
--- a/pythonlib/camoufox/server.py
+++ b/pythonlib/camoufox/server.py
@@ -63,13 +63,12 @@ def launch_server(**kwargs) -> NoReturn:
         stdin=subprocess.PIPE,
         text=True,
     )
-    # Write data to stdin and close the stream
-    if process.stdin:
-        process.stdin.write(base64.b64encode(data).decode())
-        process.stdin.close()
-
-    # Wait forever
-    process.wait()
+    # Write data to stdin, close the stream, and wait forever.
+    # communicate() tolerates the pipe closing early if the server exits before reading
+    # its config, keeping that error visible instead of masking it with an OSError.
+    process.communicate(input=base64.b64encode(data).decode())
 
     # Add an explicit return statement to satisfy the NoReturn type hint
-    raise RuntimeError("Server process terminated unexpectedly")
+    raise RuntimeError(
+        f"Server process terminated unexpectedly with exit code {process.returncode}"
+    )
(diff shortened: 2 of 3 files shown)


daijro:
Thanks for the fix!

PR #671, opened 2026-07-16
Fix getResponseBody on ORB-decompressed responses

### LINKED ISSUE #648: Network.getResponseBody fails on large compressed JSONP/script responses, but works with Accept-Encoding: identity
# Network.getResponseBody fails on large compressed JSONP/script responses, but works with Accept-Encoding: identity

## Environment

- OS: Windows
- Python package: `camoufox 0.4.11`
- Camoufox browser: `v135.0.1-beta.24`
- Playwright: `1.48.0`
- Python: `3.8`
- Browser mode: persistent context

## Problem

When listening to network responses with Playwright/Camoufox and calling `response.text()` on a large JSONP/script response, Camoufox sometimes fails with:

```text
Protocol error (Network.getResponseBody): Component returned failure code: 0x804b001b (NS_ERROR_INVALID_CONTENT_ENCODING) [nsIStreamListener.onDataAvailable]

### PR DESCRIPTION
**Pull request by pratyush618**
Fixes #648.

## Problem

`Network.getResponseBody` fails on compressed cross-origin responses with:

```
Protocol error (Network.getResponseBody): Component returned failure code:
0x804b001b (NS_ERROR_INVALID_CONTENT_ENCODING) [nsIStreamListener.onDataAvailable]
```

## Cause

Opaque Response Blocking (`browser.opaqueResponseBlocking`, enabled by default) decompresses a compressed cross-origin no-CORS response **in the parent process** in order to sniff it, unless its `Content-Type` is opaque-safelisted. It sets `hasContentDecompressed = true` but leaves `applyConversion` false.

Juggler observes the channel in the parent, so for those responses it already holds plain bytes — but `ResponseStorage.addResponseBody` infers "still compressed" from `applyConversion` alone, tags the body with its `Content-Encoding`, and `getBase64EncodedResponse` decompresses it a second time. gzip's `check_header` finds no magic bytes and throws.

The report's "large" is really a proxy for "compressed" — servers only compress above a size threshold, which is also why `Accept-Encoding: identity` worked around it.

Mozilla hit the identical bug in their own DevTools network observer — [bug 1852870](https://bugzilla.mozilla.org/show_bug.cgi?id=1852870), Firefox 120 — reported with the same error string. This change applies the same guard they did, matching the in-tree check in `NetworkResponseListener.sys.mjs`.

## Change

One additional condition in `ResponseStorage.addResponseBody`:

```js
!request.httpChannel.hasContentDecompressed
```

## Verification

Tested against the released `135.0.1-beta.24` binary (the version in the report) and the same binary with the guard applied to `chrome/juggler/content/NetworkObserver.js` inside `omni.ja`.

Matrix of 4 Content-Types × 6 Content-Encodings, each loaded as a cross-origin no-CORS script:

| | stock | patched |
|---|---|---|
| passing | 9/24 | **24/24** |

Every compressed non-safelisted combination failed before (`gzip`, `x-gzip`, `deflate`, `br`, `zstd`) — broader than the issue suggested. The `application/javascript` row passes both before and after: ORB safelists it, so Juggler still decompresses that one itself. That confirms the guard doesn't over-correct and disable legitimate decompression.

No regressions:

- `test_network.py` — 48 passed, 3 failed; the same 3 fail on the stock binary (service workers, local SSL setup)
- request-interception suites — 20/20
- fetch suites — same 7 pre-existing failures on stock and patched

## Test

`tests/async/test_network.py::test_response_text_should_work_with_compressed_cross_origin_jsonp` fails on stock with the exact reported error and passes with the fix.

It deliberately serves `application/json` rather than `application/javascript`: the latter is opaque-safelisted, skips ORB entirely, and would pass even without the fix. The existing gzip tests only cover `page.goto` navigations, which ORB never inspects — which is why this went unnoticed.

## Out of scope

- `Content-Encoding: x-deflate` also errors, identically on stock and patched. Pre-existing, unrelated, and not emitted by real servers.
- `convertString` round-trips bodies through `Array.from` + chunked `String.fromCharCode.apply`, which is inefficient for large bodies. Unchanged upstream Playwright code; separate concern.

### DIFF
--- a/additions/juggler/NetworkObserver.js
+++ b/additions/juggler/NetworkObserver.js
@@ -855,7 +855,10 @@ class ResponseStorage {
     }
     let encodings = [];
     // Note: fulfilled request comes with decoded body right away.
-    if ((request.httpChannel instanceof Ci.nsIEncodedChannel) && request.httpChannel.contentEncodings && !request.httpChannel.applyConversion && !request._fulfilled) {
+    // Note: ORB's compressed-media sniffing (browser.opaqueResponseBlocking) decodes
+    // the body in the parent process without clearing applyConversion, so
+    // hasContentDecompressed is the only signal that we already have plain bytes.
+    if ((request.httpChannel instanceof Ci.nsIEncodedChannel) && request.httpChannel.contentEncodings && !request.httpChannel.applyConversion && !request.httpChannel.hasContentDecompressed && !request._fulfilled) {
       const encodingHeader = request.httpChannel.getResponseHeader("Content-Encoding");
       encodings = encodingHeader.split(/\s*\t*,\s*\t*/);
     }
--- a/tests/async/test_network.py
+++ b/tests/async/test_network.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import asyncio
+import gzip
 import json
 from asyncio import Future
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


daijro:
Thanks for the PR ❤️

Issue #538, opened 2026-03-21
Proxy detection on FF146. WebRTC Leak


forbxpg:
## [BrowserScan](https://www.browserscan.net/) detects WebRTC for `v146.0.1-alpha.50`

If you are using a brand new Camoufox version `alpha.50` you should care about WebRTC leak when using proxies:

<img width="531" height="274" alt="Image" src="https://github.com/user-attachments/assets/ba14d8ca-1f15-478b-8f11-5e4eeb8de490" />

![Image](https://github.com/user-attachments/assets/b1d6f103-4a94-4837-81e5-9b5c2dee6064)

## For example when using Camoufox 142 from CoryKing

<img width="531" height="274" alt="Image" src="https://github.com/user-attachments/assets/df992bd8-04ce-4050-9549-33c17e6905af" />

## Version
```bash
Python Packages                                                                         
  Camoufox                        v0.5.0                                                
  Browserforge                    v1.2.4                                                
  Apify Fingerprints              v0.10.0                                               
  Playwright                      v1.58.0                                               
Browser                                                                                 
  Active                          official/prerelease/146.0.1-alpha.50                  
  Current browser                 v146.0.1-alpha.50                                     
  Installed                       Yes                                                   
  Latest in official/prerelease?  Yes                                                   
  Last Sync                       2026-03-22 02:48                                      
GeoIP                                                                                   
  Database                        MaxMind GeoLite2                                      
  Updated                         2026-03-22 02:48 
```

KemalIlkilic:
```
with Camoufox(
    geoip=True,
    proxy={
        'server': 'http://thordata.com:8080',
        'username': 'username',
        'password': 'password'
    },
    firefox_user_prefs = {
        'media.peerconnection.ice.proxy_only_if_behind_proxy': True,
        'media.peerconnection.ice.no_host': True,
    }
) as browser:
```
This should fix your issue

I had this issue earlier. I am using below code . And it works

```
from contextlib import contextmanager
from camoufox.sync_api import Camoufox
from core.config import settings


class BrowserManager:
    """Centralized browser management for consistent Camoufox configuration across all platforms"""

    _WEBRTC_LEAK_PREVENTION_PREFS = {
        'media.peerconnection.ice.proxy_only_if_behind_proxy': True,
        'media.peerconnection.ice.no_host': True,
    }

    @staticmethod
    def _build_proxy_config() -> dict | None:
        """Build proxy config dict from settings if credentials are provided"""
[...]

PR #673, opened 2026-07-16
fix: page-recycle hang under spoofed window dims (#666), + the unmerged pythonlib/settings halves of #637-#647

JWriter20:
## Addresses

- **#666** — the `new_page()` hang report: https://github.com/daijro/camoufox/pull/666#issuecomment-4995773013
- **#538** — the WebRTC / proxy-detection thread, and the beta.26 follow-up: https://github.com/daijro/camoufox/issues/538#issuecomment-4994105658

Refs #666. Refs #538. (Deliberately `Refs` rather than `Fixes` — see the UA note below; part of what #538 reports is not a Camoufox bug, so it shouldn't auto-close.)

## Overview

Two things, both surfaced while investigating reports on the `v152.0.4-beta.26` prerelease:

1. **`new_page()` hangs forever when `window.outer*` is spoofed** — the regression reported in #666. Fixed at both the browser and driver level.
2. **The WebRTC fix in beta.26 is only half of it** — the C++ half merged, the prefs half never did. Directly relevant to #538.

While tracking (2) down I found that the `pythonlib/` and `settings/` halves of #637–#647 never actually landed: when I consolidated those PRs into #666, that PR only carried `patches/` + `additions/` + one `scripts/` file. The individual PRs were closed pointing at #666, so those changes silently fell through the gap. This PR restores them.

Everything below is measured against the **official published `v152.0.4-beta.26` linux-x86_64 release**, not a local build.

---

### Root cause

The outer-size hijack in `browser-init.patch` pinned the chrome `documentElement` to the spoofed size:

```js
document.documentElement.style.setProperty('width', outerWidth + 'px');
```

That caps `.browserStack`, which caps the content viewport. Juggler's `updateViewportSize()` (`TargetRegistry.js`) then asks the content window to become Playwright's default 1280x720 — a size it can now never reach. And `awaitViewportDimensions` (`juggler/content/main.js`) awaits **exact equality with no timeout**:

```js
if (docShell.domWindow.innerWidth === width && docShell.domWindow.innerHeight === height) {
  resolve();
  return;
}
```

So non-convergence is a permanent deadlock rather than an error. The second `new_page()` hangs forever and takes the browser context with it.

**The pin was never load-bearing.** `nsGlobalWindowInner::GetOuterWidth`/`GetOuterHeight` already consult `MaskConfig` unconditionally (`fingerprint-injection.patch`), so `window.outerWidth` is spoofed in C++ regardless of the real chrome window size. Resizing is sufficient; pinning only crippled the window.

### Correcting the diagnosis in #666

The analysis in #666 blamed the inner+outer *combination* and the `if (innerWidth && innerHeight && !(outerWidth || outerHeight))` reconciliation guard. That isn't right. Measured matrix (headless, `screen 360x800`, `inner 360x740`, `outer 360x800`):

| config | before | after |
|---|---|---|
| none | pass | pass |
| `inner` only | pass | pass |
| **`outer` only** | **HANG** | pass |
| **both** | **HANG** | pass |

`outer*` **alone** is sufficient to hang, and dropping `inner*` does **not** help — so the reported "remove either group and it works" is only half true, and that guard is not the culprit. With the fix, the `both` case reports exactly `iw:360 ih:740 ow:360 oh:800`.

### Driver-side fix (works without a rebuild)

Playwright's implicit 1280x720 viewport is what asks for the impossible size, so the driver now defaults to `no_viewport` whenever the config spoofs a window dimension. Juggler then measures the window instead of resizing it, the handshake resolves immediately, and the page reports the spoofed dimensions exactly.

This matters because **it fixes the hang on already-released builds** — users get relief from PyPI without waiting for a browser rebuild. Verified against the *unpatched* official beta.26 binary: `new_page()` after `close()` returns in 0.1s reporting `iw:360 ih:740 ow:360 oh:800`. An explicit `viewport=` / `no_viewport=` from the caller always wins.

---

## 2. WebRTC: beta.26 ships the C++ half without the prefs (#538)

#666 merged `webrtc-ip-spoofing.patch` but not `settings/camoufox.cfg`, so the shipped build still has `media.peerconnection.ice.no_host=true` and **none** of the `proxy_only` prefs.

`proxy_only_if_behind_proxy` is the pref that actually stops the real-IP leak — it prevents Firefox sending a UDP STUN request around a TCP proxy, which would reveal the real IP to the STUN server. Behind a proxy no srflx forms, so the binary fabricates a synthetic srflx carrying the proxy IP (`MaybeFabricateSrflxCandidate`) and no real UDP leaves the box. Without the pref, the C++ sanitizing is doing less than intended.

`no_host=false` keeps the stock two-candidate shape (an mDNS `<uuid>.local` host + a srflx); `obfuscate_host_addresses` makes that leak-free. `no_host=true` leaves a single-srflx shape that doesn't match a real browser — a tell in its own right.

This may well be why #538 reporters still see blocking on beta.26.

> **Note on the "Firefox 152 vs 150" report in #538:** I could not reproduce that as a Camoufox bug. On the official beta.26 the driver correctly injects `Firefox/152.0` (`rv:152.0`) and JS `navigator.userAgent` matches the HTTP `User-Agent` header exactly. Launched *without* the driver, though, the UA is `...rv:152.0) Gecko/20100101 Camoufox/152.0.4-beta.26` — no `Firefox/` token at all. That plus unspoofed WebRTC matches the reported symptoms, so I believe those reports are from launching the binary directly rather than through the launcher.

---

## 3. Restored from the #666 consolidation

- **`fix(proxy)`** (#638) — `dom.security.https_first` rewrote `http://` to `https://` before the launch-arg proxy filter saw it, breaking CONNECT-only proxies. Stock Playwright Firefox builds disable this.
- **`fix(stealth)`** (#646) — speech-voice spoofing didn't apply, and host voices leaked.
- **`fix(stealth)`** (#647) — clamp `inner <= outer <= avail <= screen` on both axes. BrowserForge can emit impossible geometries (e.g. `outerWidth > screen.width`), which the browser faithfully reports and which leak as tells.

---

## Testing

- `pythonlib/tests/` — 43 passing, including new coverage for the `no_viewport` default (explicit caller choice wins; config-chunk reassembly is index-ordered, so `CAMOU_CONFIG_10` can't sort before `_2`), voices, and geometry clamping.
- The #666 matrix above was run end-to-end against the official beta.26 binary, before and after.
- These are **runtime-only** regressions — the patch applies cleanly either way, so only driving a real browser catches them. Worth noting given #666's screenshot/popup breakage had the same property.

## Known limitation

With `outer*` set but `inner*` unset, the page now reports `iw:500 > ow:360` — an impossible geometry. It's strictly better than the hang it replaces, and the driver never produces that combination (it always pairs inner+outer and clamps), but a raw `CAMOU_CONFIG` user could still hit it. Happy to follow up if you'd like the browser to derive an inner cap from `outer*`.


KittisakPh:
@daijro this look good 😊

daijro:
Thanks for the quick fix!

JWriter20:
> Thanks for the quick fix!

Glad to help! Think everything is up-to-date with my fork now and I had been using that build for a little while before I switched over to something else, so it should be fairly stable now.

KittisakPh:
@JWriter20 “ before I switched over to something else” What you mean ???

JWriter20:
> @JWriter20 “ before I switched over to something else” What you mean ???

My main repo I am working on that used camoufox is in Typescript, so I had to constantly jump between changes in the camoufox-js and camoufox repo. Got tired of that so I made my own private version I am now maintaining and improving. I will still help with little fixes in this repo occasionally but most of my development will now be on that version and my business that uses it.

KittisakPh:
@JWriter20 Could you explain what the current main Camoufox project is still missing—the things that made a personal project necessary? I don't quite understand, because if everyone contributed to the main project, development would move much faster.

JWriter20:
> @JWriter20 Could you explain what the current main Camoufox project is still missing—the things that made a personal project necessary? I don't quite understand, because if everyone contributed to the main project, development would move much faster.

Having to jump between two projects in different languages that are not maintained by me was the biggest issue. But there are also some subtle stealth issues in this repo that require a lot of effort to fix that I need for my use case. If I were to release the fixes, then it would be easy for captcha companies to reverse engineer and find ways to detect my new spoofs. Because my business is relying on this stealth to a degree, I have to make most of my new contributions private. If I end up switching over to a chromium based solution I will probably make it public, but otherwise its not worth it.  

The good news for people using this project though, is that basically all of the issues come from cross-OS emulation, @KittisakPh since you are using windows, this is likely not an issue for you. Just make sure the fingerprints you are using are also windows and you shouldn't have issues.

introvert:
Good job, testing this @JWriter20

JWriter20:
> Good job, testing this @JWriter20

Thank you, would appreciate that

PR #674, opened 2026-07-17
Clamp headful window geometry to the real display

### LINKED ISSUE #499: Window rendering defect. Flickering effect on right and bottom margins.
When I open the browser, the screen doesn't render correctly. Sometimes it's larger than the actual screen monitor. Other times, there are visual defects on the right and bottom margins.
This happens with any website.

Example code:

	# -*- coding: utf-8 -*-
	import asyncio

	async def main():
	    async with AsyncCamoufox(headless=False) as browser:
	        await page.goto("https://www.google.com/", wait_until="domcontentloaded")
	        await page.wait_for_timeout(3000)
	        pass

	asyncio.run(main())

Technical data:
	
	>>> Pip package:    v0.4.11

	>>> Distributor ID:	Ubuntu
		Description:	Ubuntu 24.04.4 LTS
		Release:	24.04
		Codename:	noble

	>>> amd64

	>>> Intel Corporation TigerLake-LP GT2 [Iris Xe Graphics] (rev 01)

	>>> Screen 0: minimum 320 x 200, current 1366 x 768, maximum 16384 x 16384
		eDP-1 connected primary 1366x768+0+0 (normal left inverted right x axis y axis)

Example image:

Image

pratyush618:
Fixes #499.

## Problem

In headful mode the browser window is sometimes larger than the monitor, with rendering artifacts along the right and bottom margins. Reported on a 1366x768 laptop.

## Cause

`get_screen_cons()` already computes the right bound — `Screen(max_width=1366, max_height=768)` — and hands it to BrowserForge. BrowserForge then throws it away.

`FingerprintGenerator.partial_csp` filters the fingerprint pool by the screen constraint, and when that filtering fails it swallows the error and drops the constraint entirely unless `strict=True`:

```python
try:
    return get_possible_values(self.fingerprint_generator_network, filtered_values)
except Exception as e:
    if strict:
        raise e
    del filtered_values['screen']
return None
```

Camoufox doesn't pass `strict`, so the bound is silently discarded and a 1366x768 laptop gets a 2560x1440 fingerprint with `window.outerWidth` 1920. `browser-init.patch` then calls `window.resizeTo(outerWidth, outerHeight)` on the real chrome window, so the oversized value becomes a real oversized window.

Measured on an Xvfb sized to the reporter's 1366x768, the real X window reached **2225x1238**.

## Change

Re-apply the bound after generation rather than trusting BrowserForge with it:

- `clamp_screen_to_display()` — shrinks `screen.width/height` to the display, preserving the taskbar delta so `fix_screen_no_taskbar`'s `avail < height` invariant still holds. The existing `clamp_window_dimensions()` then cascades the new bounds down to avail/outer/inner.
- `clamp_window_position()` — pulls `screenX/screenY` back inside the screen. Shrinking the screen invalidates the positions BrowserForge generated against the original one, which would otherwise leave the window box partly outside its own reported screen.

## Headful only

The clamp is gated on `headless is False`, which matters more than it looks:

- headless has no window to overflow.
- `headless='virtual'` launches Xvfb with `-screen 0 1x1x24` (`virtdisplay.py`), and `env` *is* `os.environ`, so `env['DISPLAY'] = virtual_display` mutates the caller's environment. Under that display `get_screen_cons(True)` returns `Screen(max_width=1, max_height=1)` — verified. An unconditional clamp would shrink those fingerprints to a 1x1 screen. It is harmless today only because BrowserForge discards the constraint.

## Verification

Against the released binary on an Xvfb at 1366x768:

| | before | after |
|---|---|---|
| generated geometry fits the display | 7/40 | **40/40** |
| full geometry invariants hold | 17/60 | **50/50** |

Invariants checked: `inner <= outer <= avail <= screen`, `screen <= display`, `avail < screen` (noTaskbar tell), all dimensions positive, and `0 <= screenX/Y <= screen - outer`.

Real window after the fix: visible chrome **1366x720 at +5+5**, inside the display. The toplevel X window reads 1376x730 because Firefox carries a 5px invisible shadow border on each side (there is no window manager under Xvfb, and no `_GTK_FRAME_EXTENTS`); the child window matches `window.outerWidth` exactly.

Headless and `headless='virtual'` are unaffected — sampled under the 1x1 Xvfb they still produce 3072x1728 and 3440x1440, so no fingerprint diversity is lost.

## Tests

`pythonlib/tests/test_fingerprint_fixes.py`, 12 -> 20 passing. Covers the shrink, no-op when already within the display, unset bounds, the avail floor, composition with `clamp_window_dimensions` (the ordering is load-bearing), the position clamp, and the negative guard.

Full pythonlib suite: 53 passed. The 2 `test_virtdisplay.py` failures are pre-existing and reproduce unchanged on a clean tree.

## Not addressed

BrowserForge silently dropping non-strict constraints is an upstream bug. This change defends against it rather than depending on it being fixed.

### DIFF
--- a/pythonlib/camoufox/fingerprints.py
+++ b/pythonlib/camoufox/fingerprints.py
@@ -406,6 +406,52 @@ def clamp_window_dimensions(config: Dict[str, Any]) -> None:
             config[f'window.inner{axis}'] = outer_clamped
 
 
+def clamp_screen_to_display(
+    config: Dict[str, Any],
+    max_width: Optional[int],
+    max_height: Optional[int],
+) -> None:
+    """Shrink screen.width/height down to the bounds of the real display.
+
+    BrowserForge takes a Screen constraint but drops it silently whenever it
+    filters the fingerprint pool too far: FingerprintGenerator.partial_csp
+    swallows the resulting failure and deletes the constraint unless strict=True.
+    So the bound from get_screen_cons() is best-effort only, and a 1366x768
+    laptop routinely gets a 2560x1440 fingerprint. browser-init.patch resizes the
+    real chrome window to window.outerWidth/outerHeight, so an unbounded value
+    renders past the edge of the monitor (daijro/camoufox#499).
+
+    Keeps the taskbar delta (screen - avail) intact so fix_screen_no_taskbar's
+    invariant survives. Callers must run clamp_window_dimensions afterwards to
+    cascade the new bounds down to avail/outer/inner.
+    """
+    for axis, cap in (('width', max_width), ('height', max_height)):
+        screen = config.get(f'screen.{axis}')
+        if not (screen and cap) or screen <= cap:
+            continue
+        avail_key = 'screen.availWidth' if axis == 'width' else 'screen.availHeight'
+        avail = config.get(avail_key)
+        config[f'screen.{axis}'] = cap
+        if avail:
+            config[avail_key] = max(1, cap - max(0, screen - avail))
+
+
+def clamp_window_position(config: Dict[str, Any]) -> None:
+    """Keep the window box inside the screen: 0 <= screenX/Y <= screen - outer.
+
+    BrowserForge's screenX/screenY are consistent with the screen it generated
+    them against, so clamp_screen_to_display invalidates them. A window
+    positioned partly off its own reported screen is an impossible geometry.
+    """
+    for axis, pos_key in (('Width', 'window.screenX'), ('Height', 'window.screenY')):
+        screen = config.get(f'screen.{axis.lower()}')
+        outer = config.get(f'window.outer{axis}')
+        pos = config.get(pos_key)
+        if pos is None or not (screen and outer):
+            continue
+        config[pos_key] = max(0, min(pos, screen - outer))
+
+
 def set_media_devices_defaults(config: Dict[str, Any]) -> None:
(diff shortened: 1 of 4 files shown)


pratyush618:
Follow-up pushed: `headless='virtual'` reaches `launch_options` as `headless=False` with `virtual_display` set (`async_api.py` rewrites it before the call), so the `headless is False` gate fired for it after all.

Under the 1x1 Xvfb the clamp shrank `screen.height` to 1, `fix_screen_no_taskbar` then subtracted the 40px taskbar, and `validate_config` rejected the launch:

```
camoufox.exceptions.InvalidPropertyType: Invalid type for property screen.availHeight. Expected uint, got int
```

Gate now also excludes `virtual_display`, with a launch-level regression test in `tests/test_launch_geometry.py` (fails with the exception above without the fix).

#694 is stacked on this branch.

JWriter20:
Your work landed in `v152.0.4-beta.29` via #707, as commits `2266f27` ("Clamp headful window geometry to the real display"), `22c6ffb` ("Probe the host monitor in CSS pixels"), `fbafbcf` ("Exclude virtual displays from the display clamp") and `a5afa46` ("Apply screen constraints on Windows and macOS"). This closed #499.

Closing as superseded — thank you, this was a genuinely useful fix.

PR #679, opened 2026-07-17
fix(juggler): restore humanized mouse trajectory dropped in FF146 migration (#677)

### LINKED ISSUE #677: humanize=True emits only endpoint mousemove on Linux since Juggler FF146 migration
### Describe the bug:

Camoufox's native mouse humanization has no effect on Linux with browser
v152.0.4-beta.27 and Python package v0.5.4.

With `humanize=True`, a long `page.mouse.move()` emits only one `mousemove`
event at the destination. There are no intermediate trajectory points.

The same result occurs with `humanize=1.5`.

This affects normal `locator.click()` calls as well as drag and press-and-hold
interactions used by CAPTCHA solvers.

The documented behavior says `humanize` enables Camoufox's native C++
human-like cursor trajectory:

https://camoufox.com/fingerprint/cursor-movement/

### Source investigation:

The C++ trajectory generator still exists:

`additions/camoucfg/MouseTrajectories.hpp`

The current v152 `PageHandler.js`, however, dispatches the requested mousemove
directly and does not call `ChromeUtils.camouGetMouseTrajectory()`:

https://github.com/CloverLabsAI/camoufox/blob/v152.0.4-beta.27/additions/juggler/protocol/PageHandler.js

The trajectory integration previously existed in `PageHandler.js`, but appears
to have been removed by commit `03c1230` during the Firefox 146 Juggler
migration:

https://github.com/CloverLabsAI/camoufox/commit/03c12302b0ee248230dbf26e929937da7ca103f4

That commit removed the branch which:

1. Checked `ChromeUtils.camouGetBool("humanize", false)`.
2. Called `ChromeUtils.camouGetMouseTrajectory(...)`.
3. Dispatched each intermediate trajectory point with a short delay.

The call remains absent in the FF146, v150, and v152 tags, even though the
generator and configuration properties remain present.

### Screenshots:

Not required. The emitted DOM events demonstrate the issue.

### To Reproduce:

Run inside Linux:

```python

with Camoufox(
    headless=True,
    humanize=True,
    os="linux",
) as browser:
    page = browser.new_page()
    page.set_content(
        '<body style="margin:0;width:1400px;height:800px"></body>'
    )

    page.evaluate("""
        window.moves = [];
        addEventListener("mousemove", event => {
            moves.push([event.clientX, event.clientY]);
        });
    """)

    page.mouse.move(20, 20)
    page.mouse.move(1100, 650)

    print(page.evaluate("moves"))
```

Actual output:

```text
[[20, 20], [1100, 650]]
```

Expected output:

Multiple intermediate `mousemove` coordinates forming a humanized trajectory
between `[20, 20]` and `[1100, 650]`.

### Version:

```text
Python Packages
  Camoufox                    v0.5.4
  Browserforge                v1.2.4
  Apify Fingerprints          v0.13.0
  Playwright                  v1.60.0
Browser
  Active                      official/stable
  Current browser             v152.0.4-beta.27
  Build date                  Jul 16
  SHA256                      5fd73824da8f
  Installed                   Yes
  Latest in official/stable?  Yes
  Last Sync                   2026-07-17 10:44
GeoIP
  Status                      Not supported (install camoufox[geoip])
Storage
  Install path                /root/.cache/camoufox
  Browser(s) directory size   1.2 GB
  Config file                 /root/.cache/camoufox/config.json
  Repo cache                  /root/.cache/camoufox/repo_cache.json
```

Additional environment details:

- Debian GNU/Linux 12 (bookworm)
- Linux ARM64
- Python 3.11.2

It is also reproducible on macOS ARM64, but this report specifically confirms
that the behavior is not limited to the existing macOS issue:

https://github.com/daijro/camoufox/issues/498

JWriter20:
This was actually a fairly minor fix, but seems like its been around awhile. Just affects juggler.
__________________________________________________________________

Fixes #677

### Problem

Since the Firefox 146 Juggler migration (commit `03c1230`), `humanize=True` has had no effect: a long `page.mouse.move()` emits only the single endpoint `mousemove`, with no intermediate trajectory points. This affects `locator.click()`, drag, and press-and-hold flows used by CAPTCHA solvers.

As the issue correctly diagnosed, `03c1230` rewrote the `sendEvents()` helper inside `Page.dispatchMouseEvent` to batch events via `Promise.all`, and in doing so dropped the Camoufox humanize branch. The native C++ generator (`MouseTrajectories.hpp` / `ChromeUtils.camouGetMouseTrajectory`) and all config plumbing (`settings/properties.json`, `pythonlib`) remained intact — only the call site in `PageHandler.js` was lost.

### Fix

Reintroduce the trajectory expansion inside the current `sendEvents()`:

- Factor single-event dispatch into a `sendOne()` helper (keeps the FF152 `jugglerSendMouseEvent` signature).
- When `humanize` is enabled, expand a `mousemove` into the C++-generated trajectory, dispatching intermediate points sequentially with a 10 ms delay, then **always finishing exactly on the requested destination** so clicks/hover still land on target (the raw trajectory's last point is skipped, matching the original loop, but the exact endpoint is now dispatched explicitly for accuracy).
- Restore `_lastTrackedPos` so each move's trajectory starts from the real previous cursor position.

### Regression test

Adds `tests/patches/humanize-mouse-trajectory.py` (following the existing `tests/patches/` convention) that drives a real build and counts emitted mousemove events:

- `humanize=True` must expand a long move into ≥10 intermediate points ending exactly on the destination;
- a humanized `click()` must still land on the target element;
- without humanize, each move emits only its endpoint (pins the other direction so accidental always-on humanization is caught too).

### Verification

Tested end-to-end against the built `v152.0.4-beta.25` Linux binary by swapping the updated loose `PageHandler.js` into the dist (no rebuild needed). The regression test was confirmed to have teeth — it **fails on the pre-fix code and passes on the fix**:

| Config | pre-fix | with fix |
|---|---|---|
| `humanize=True`, long move | 2 events (bug) | ~149 events, ends `[1100,650]` |
| humanized `click()` | 1 intermediate move (missed humanization) | ~111 moves, lands on target |
| `humanize` off | 2 endpoints | 2 endpoints (unchanged) |
| test exit code | `1` (FAIL) | `0` (PASS) |

Regression checks on the same binary: normal `click`/`fill`/`drag_and_drop` behave identically.

### DIFF
--- a/additions/juggler/protocol/PageHandler.js
+++ b/additions/juggler/protocol/PageHandler.js
@@ -81,6 +81,9 @@ export class PageHandler {
 
     this._isDragging = false;
     this._lastMousePosition = { x: 0, y: 0 };
+    // Camoufox: last position an actual mousemove landed on. Used as the
+    // start point for the humanized cursor trajectory (humanize=True).
+    this._lastTrackedPos = { x: 0, y: 0 };
 
     this._reportedFrameIds = new Set();
     this._networkEventsForUnreportedFrameIds = new Map();
@@ -523,13 +526,14 @@ export class PageHandler {
         await helper.awaitTopic('apz-repaints-flushed');
 
       const watcher = new EventWatcher(this._pageEventSink, types, this._pendingEventWatchers);
-      const promises = [];
-      for (const type of types) {
+      // Dispatch a single synthesized mouse event to the renderer and return a
+      // promise that resolves once the renderer acks it.
+      const sendOne = (eventType, eventX, eventY) => {
         // This dispatches to the renderer synchronously.
         const jugglerEventId = win.windowUtils.jugglerSendMouseEvent(
-          type,
-          x + boundingBox.left,
-          y + boundingBox.top,
+          eventType,
+          eventX + boundingBox.left,
+          eventY + boundingBox.top,
           button,
           clickCount,
           modifiers,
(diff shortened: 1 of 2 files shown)


JWriter20:
@daijro Pr with fix with regression test

PR #692, opened 2026-07-21
tests(patches): pin main-world evaluation of statements (#631)

### LINKED ISSUE #631: if(){} statement in js breaks main_world_eval
### Describe the bug:
using main_world_eval to execute arbitrary js in the main world will result in an "Execution context was destroyed" error if the js contains an "if" statement. The ternary operator, however, causes no issues. I'm not much of a js developer... maybe it has something to do with closures?

On Python 3.12, using camoufox 0.4.11

### To Reproduce:
MVE:
```
import asyncio
from camoufox import async_api as cf_async_api


async def main():
    async with cf_async_api.AsyncCamoufox(headless=False, main_world_eval=False) as browser_1:
        page_1 = await browser_1.new_page()
        await page_1.goto("https://httpbin.org", wait_until='load')
        result_1 = await page_1.evaluate("{foo: 'bar'}")
        print(result_1)
        result_2 = await page_1.evaluate("if(true){{foo: 'bar'}}else{{foo: 'baz'}}")
        print(result_2)

    async with cf_async_api.AsyncCamoufox(headless=False, main_world_eval=True) as browser_2:
        page_2 = await browser_2.new_page()
        await page_2.goto("https://httpbin.org", wait_until='load')
        result_3 = await page_2.evaluate("mw:{foo: 'bar'}")
        print(result_3)
        result_4 = await page_2.evaluate("mw:true ? {foo: 'bar'} : {foo: 'baz'}")
        print(result_4)
        result_5 = await page_2.evaluate("mw:if(true){{foo: 'bar'}}else{{foo: 'baz'}}")
        print(result_5)

if __name__ == '__main__':
    asyncio.run(main())
```

Also worth noting the output is different depending on main_world_eval true/false, which is counterintuitive (separate issue)
### Version:

Run `python -m camoufox version` in your terminal and paste the output here.
Pip package:    v0.4.11

neuregex:
## Related Issue

Closes #631

## Description

Camoufox evaluates page scripts through its own main-world path (the `main_world_eval` launch flag and the Juggler evaluation handler) rather than Playwright's default isolated world. That path has to wrap whatever string it is handed so a bare **statement** — `if (...) { ... }`, `let x = 1;`, a block — still produces a value instead of erroring out.

#631 reported that any script containing an `if` statement failed with `Execution context was destroyed` on camoufox 0.4.11, while the equivalent ternary worked. **That symptom is gone on current builds** (report below), so this PR adds a regression test rather than a fix — the issue looks resolved somewhere between 0.4.11 and 0.5.4, and this pins it so a future Juggler rebase can't silently reintroduce it. The failure mode is a hard error on the caller's side that nothing in the build catches, which is exactly the class of regression `tests/patches/` exists for.

Follows the existing convention in that directory (standalone script, `main()` returning 0/1, `CAMOUFOX_EXECUTABLE_PATH` support, positive cases in both modes) — same shape as `humanize-mouse-trajectory.py`.

If you'd rather keep #631 open until the original reporter confirms, feel free to drop the `Closes` — the test stands on its own either way.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Other — test coverage only (no patches, browser source or `pythonlib/` behaviour touched)

## Testing

Verified against **v152.0.4-beta.25** (pythonlib 0.5.4, Python 3.12, Windows):

```
$ CAMOUFOX_EXECUTABLE_PATH=/path/to/camoufox.exe python tests/patches/main-world-eval.py

=== main_world_eval=False ===
  PASS object literal       -> {'foo': 'bar'}
  PASS ternary              -> {'foo': 'bar'}
  PASS if statement         -> {'foo': 'bar'}
  PASS if numeric           -> 1
  PASS let + if             -> 2
  PASS arrow IIFE with if   -> 1

=== main_world_eval=True ===
  PASS object literal       -> {'foo': 'bar'}
  PASS ternary              -> {'foo': 'bar'}
  PASS if statement         -> {'foo': 'bar'}
  PASS if numeric           -> 1
  PASS let + if             -> 2
  PASS arrow IIFE with if   -> 1

PASS: statements and expressions evaluate identically in both modes

$ echo $?
0
```

## Fingerprint Report

This PR adds a single test file and changes no patches, no browser source and no `pythonlib/` behaviour, so there is no fingerprint surface for it to affect. Happy to run both suites anyway if you'd like them attached.

### DIFF
--- a/tests/patches/main-world-eval.py
+++ b/tests/patches/main-world-eval.py
@@ -1,0 +1,93 @@
+"""
+Verify main-world evaluation accepts statements, not just expressions (daijro/camoufox#631).
+
+Camoufox evaluates page scripts through its own main-world path (the
+`main_world_eval` launch flag and the juggler evaluation handler) rather than
+Playwright's default isolated world. That path has to wrap whatever string it
+is handed so a bare *statement* -- `if (...) { ... }`, `let x = 1;`, a block --
+still produces a value instead of a syntax error.
+
+#631 reported that any script containing an `if` statement failed with
+"Execution context was destroyed" on camoufox 0.4.11, while the equivalent
+ternary worked. That symptom is gone on current builds, so this test pins the
+behaviour: the wrapping is easy to break again during a Juggler rebase, and the
+failure mode is a hard error on the caller's side rather than anything the
+build catches.
+
+Run against a specific build:
+    CAMOUFOX_EXECUTABLE_PATH=/path/to/camoufox-bin python tests/patches/main-world-eval.py
+(without the env var it uses the camoufox-managed browser download.)
+
+What PASS means:
+    * expressions (object literal, ternary, arrow IIFE) evaluate to their value;
+    * statements (`if`, `let` + reassignment, block) evaluate without raising
+      and return the expected completion value;
+    * both main_world_eval=True and main_world_eval=False behave identically,
+      so neither path regresses independently.
+"""
+
+import asyncio
+import os
+import sys
+from typing import Any, Dict, List, Tuple
+
+from camoufox.async_api import AsyncCamoufox
+
+EXECUTABLE_PATH = os.environ.get("CAMOUFOX_EXECUTABLE_PATH")
+
+# (label, script, expected)
+CASES: List[Tuple[str, str, Any]] = [
+    ("object literal", "({foo: 'bar'})", {"foo": "bar"}),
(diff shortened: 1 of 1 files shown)


JWriter20:
Your tests landed in `v152.0.4-beta.29` via #707, as commit `35c133c` ("tests(patches): pin main-world evaluation of statements (#631)").

Closing as superseded — thanks for pinning this down.

PR #694, opened 2026-07-21
Fix window opening larger than the screen under display scaling

### LINKED ISSUE #425: Window opens larger than screen with display scaling enabled
Camoufox opens a window that extends beyond screen boundaries when Windows display scaling is enabled (150%/200%).

Steps to Reproduce:

Enable display scaling in Windows 11 (e.g., 150%)
Launch Camoufox
Window opens larger than visible screen space

Expected: Window should fit within screen bounds accounting for scaling
Actual: Window exceeds physical screen dimensions

pratyush618:
Fixes #425.

> **Stacked on #674.** Only the last two commits belong to this PR:
> `Probe the host monitor in CSS pixels` and `Apply screen constraints on Windows and macOS`.
> The first two are #674's branch, which this builds on to avoid duplicating its clamp.
> Rebase onto `main` once #674 lands.

## Problem

With Windows display scaling enabled (150% / 200%), the browser window opens larger than the monitor.

## Cause

Two independent defects, both in `get_screen_cons()`'s contract.

**1. The bound is computed in the wrong unit.**

`screeninfo`'s Windows enumerator calls `SetProcessDpiAwareness(2)` before enumerating, explicitly so that it reports *physical* pixels:

```python
# Make the process DPI aware so it will detect the actual
# resolution and not a virtualized resolution reported by
# Windows when DPI virtualization is in use.
ctypes.windll.shcore.SetProcessDpiAwareness(2)
```

Firefox lays windows out in **CSS pixels**. Nothing in this repo pins `layout.css.devPixelsPerPx`, so it stays at `-1` and tracks the system scale factor. `browser-init.patch` then calls `window.resizeTo(outerWidth, outerHeight)` with a value the Python layer derived from physical pixels.

At 150% on a 1920x1080 panel the CSS screen is 1280x720, so a fingerprint bounded at "1920x1080" produces a window of 2880x1620 device pixels — 1.5x the screen on both axes.

**2. The bound is never applied on Windows or macOS.**

```python
screen_cons = screen or get_screen_cons(headless or 'DISPLAY' in env)
```

`DISPLAY` is X11-only. On Windows and macOS it is never set, so headful runs took the `headless is False` early return and generated with **no monitor bound at all** — the scaling mismatch above then applied to whatever BrowserForge picked, up to 2560x1440.

## Change

**`camoufox/display.py`** (new) — probes the host monitor in CSS pixels.

- `largest_display()` returns a `DisplaySize` in CSS px. On Windows it divides the physical size by the monitor's effective DPI (`shcore!GetDpiForMonitor` via `MonitorFromPoint`); macOS (`NSScreen.frame`) and X11 already report CSS px, so scaling is Windows-only.
- Degrades to 1.0 whenever the DPI cannot be read (pre-Windows 8.1, `shcore` missing, non-zero `HRESULT`), so behaviour is unchanged where it does not apply.
- Uses private `ctypes.WinDLL` handles rather than the process-wide `ctypes.windll` cache, so annotating the prototypes cannot leak into other libraries in the same process.
- `has_display(env)` replaces the `'DISPLAY' in env` probe: always true off Linux, `DISPLAY` or `WAYLAND_DISPLAY` on Linux.

`get_screen_cons()` keeps its signature and now returns CSS-pixel bounds.

## Scope

Untouched: headless (no window to overflow), `headless='virtual'` (excluded by #674's gate), any caller passing `screen=` or `window=`, and any host whose display cannot be probed.

Windows without scaling is also unaffected — `GetDpiForMonitor` returns 96 DPI, the scale factor is 1.0, and the bound is the physical size as before.

## Tests

`pythonlib/tests/test_display.py` (new) — the DPI conversion table (96/120/144/192 -> 1.0/1.25/1.5/2.0), the Windows-only guard, both fallback paths, monitor selection, and `has_display` per platform.

`pythonlib/tests/test_launch_geometry.py` — adds `TestHeadfulFitsOnDisplay`: 15 randomised headful launches against a 1280x720 CSS display, asserting no `screen.*` or `window.outer*` value exceeds it, plus the geometry invariants and the unprobeable-host case. Asserts the outcome rather than the constraint, since BrowserForge silently drops constraints it cannot satisfy.

Full pythonlib suite: **89 passed**. The 2 `test_virtdisplay.py` failures are pre-existing (no Xvfb on this host) and reproduce unchanged on a clean tree.

## Not addressed

Launching the binary directly, without the Python package, still uses the hardcoded `window.resizeTo(1280, 1040)` in `browser-init.patch`, which can overflow a small scaled display. That needs a patch regeneration through `make edits`.

### DIFF
--- a/pythonlib/camoufox/display.py
+++ b/pythonlib/camoufox/display.py
@@ -1,0 +1,101 @@
+"""
+Host display geometry, in the units Firefox lays its windows out in.
+
+Firefox sizes windows in **CSS pixels**, but `screeninfo` marks the process
+per-monitor DPI aware and therefore reports **physical** pixels. Where Windows
+display scaling is enabled the two differ by the scale factor: a 1920x1080 panel
+at 150% is only 1280x720 CSS px. Deriving a window size from the physical
+numbers overshoots the screen by that factor, so the window opens partly
+off-screen (daijro/camoufox#425).
+
+macOS (`NSScreen.frame`) and X11 (xrandr) already report CSS pixels, so scaling
+only ever applies on Windows.
+"""
+
+from typing import Any, NamedTuple, Optional
+
+from screeninfo import get_monitors
+
+from .pkgman import OS_NAME
+
+# Windows expresses DPI relative to this baseline: 144 DPI == 150% scaling.
+_WINDOWS_BASE_DPI = 96
+
+# shcore.h / winuser.h constants
+_MDT_EFFECTIVE_DPI = 0
+_MONITOR_DEFAULTTONEAREST = 2
+
+
+class DisplaySize(NamedTuple):
+    """Size of a monitor in CSS pixels."""
+
+    width: int
+    height: int
+
+
+def largest_display() -> Optional[DisplaySize]:
+    """
+    Size of the roomiest attached monitor in CSS pixels, or None when the
+    display cannot be probed (no monitors, or enumeration failed).
+    """
+    try:
+        monitors = get_monitors()
+    except Exception:
+        return None
+    if not monitors:
+        return None
+
+    monitor = max(monitors, key=lambda m: m.width * m.height)
+    scale = _scale_factor(monitor)
+    return DisplaySize(
+        width=max(1, int(monitor.width / scale)),
+        height=max(1, int(monitor.height / scale)),
+    )
+
+
+def _scale_factor(monitor: Any) -> float:
+    """
+    Physical pixels per CSS pixel on `monitor`. Always 1.0 outside Windows.
+    """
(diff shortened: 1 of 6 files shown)


DemonMartin:
When will this be shipped? I was forced to do a very hacky fix myself for now but it was causing flags with a lot of captchas

JWriter20:
Your work landed in `v152.0.4-beta.29` via #707, through the display-clamp series — `2266f27` ("Clamp headful window geometry to the real display") and `a5afa46` ("Apply screen constraints on Windows and macOS"), with `22c6ffb` probing the host monitor in CSS pixels. This closed #425.

PR #697, opened 2026-07-22
Spoof navigator.maxTouchPoints from config

### LINKED ISSUE #696: Spoof navigator.maxTouchPoints from config
### Use case

`navigator.maxTouchPoints` is already a recognized config property — `settings/properties.json` declares `{ "property": "navigator.maxTouchPoints", "type": "uint" }` — but **nothing reads it**. It is validated and then silently ignored, so it always reports `0` on a desktop build regardless of what the config sets.

Under a Firefox-for-Android user agent this is a mobile-detection tell: a real phone reports `maxTouchPoints = 5`, but Camoufox reports `0`. A desktop build has no touch digitizer, and Firefox's RFP path only ever *collapses* the value to `0` (`MaxTouchPointsCollapse`) — there is no path to a phone value without an explicit override.

Verified live: setting `navigator.maxTouchPoints` in `CAMOU_CONFIG` currently has no effect (stays `0`); disabling the RFP collapse target and enabling touch-events prefs also has no effect, because the underlying host value is genuinely `0`.

### Why it's useful

`maxTouchPoints` is the one touch signal that needs a browser-level override. The rest of the mobile touch surface (touch events, `pointer: coarse`, `hover: none`) is already reachable via Playwright `hasTouch`, so honouring the existing config property closes the remaining gap and makes a consistent Android fingerprint possible.

### Proposed change

Read the already-declared config property in `Navigator::MaxTouchPoints` via the standard `MaskConfig` early-return, matching how every other navigator/screen property is spoofed. No schema change required.

nxbdev:
## Related Issue

Closes #696

## Description

`navigator.maxTouchPoints` was already a recognized config property (`settings/properties.json` declares it) but nothing read it, so it always reported `0` on a desktop build regardless of the config. Under a Firefox-for-Android user agent that `0` is a mobile-detection tell — a real phone reports `5`, and Firefox's RFP path only ever collapses the value to `0` (`MaxTouchPointsCollapse`), with no path to a phone value without an explicit override.

This adds the standard `MaskConfig` early-return at the top of `Navigator::MaxTouchPoints`, matching how every other navigator/screen property is spoofed (`GetPlatform`, `GetOscpu`, `HardwareConcurrency`, `nsScreen::PixelDepth`, …):

```cpp
if (auto value = MaskConfig::GetUint32("navigator.maxTouchPoints")) {
  return value.value();
}
```

`Navigator.cpp` already includes `MaskConfig.hpp` via `navigator-spoofing.patch`, so no extra include is needed and no schema change is required. With this, `navigator.maxTouchPoints` in the config is honoured (e.g. `5` for a phone).

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Other

## Testing

Built and driven against the raw binary (same mechanism as `build-tester/runner.py`: `firefox.launch(executable_path=…, env={CAMOU_CONFIG})`), reading `navigator.maxTouchPoints` on a page:

| Case | `CAMOU_CONFIG` | Result |
| --- | --- | --- |
| Control (no config) | — | `0` (unchanged desktop default) |
| Phone value | `{"navigator.maxTouchPoints": 5}` | `5` |
| Single-touch | `{"navigator.maxTouchPoints": 1}` | `1` |

- Patch applies cleanly against the current `main` (Firefox `152.0.4`, `beta.28`) — `patching file dom/base/Navigator.cpp`, no rejects/fuzz.
- Verified on a `linux/arm64` build.

## Fingerprint Report

<details>

_build-tester score/screenshot to be attached once the current beta.28 arm64 build completes — this PR is opened as a draft until then._

</details>

## Checklist

- [ ] ~~Service tests pass~~ (no `pythonlib/` changes)
- [ ] Build test passes (score ≥ 1000, screenshot) — pending, will attach before marking ready

### DIFF
--- a/patches/mobile-fingerprint-spoofing.patch
+++ b/patches/mobile-fingerprint-spoofing.patch
@@ -1,0 +1,17 @@
+diff --git a/dom/base/Navigator.cpp b/dom/base/Navigator.cpp
+--- a/dom/base/Navigator.cpp
++++ b/dom/base/Navigator.cpp
+@@ -896,6 +896,13 @@
+ //*****************************************************************************
+
+ uint32_t Navigator::MaxTouchPoints(CallerType aCallerType) {
++  // Camoufox: allow spoofing navigator.maxTouchPoints via config. A desktop
++  // build has no touch digitizer, so the real value is 0 and Firefox's RFP
++  // path only ever collapses it to 0 - there is no way to report a phone's
++  // value without this override.
++  if (auto value = MaskConfig::GetUint32("navigator.maxTouchPoints")) {
++    return value.value();
++  }
+   nsIDocShell* docshell = GetDocShell();
+   BrowsingContext* bc = docshell ? docshell->GetBrowsingContext() : nullptr;
+


JWriter20:
Thanks @nxbdev. Landed via #756 (merged as 35b45c1), commit 5361844, authorship intact.

Built on top of it in a80abb4, because `navigator.maxTouchPoints` alone was self-contradicting: `(any-pointer: coarse)` stayed false and `window.TouchEvent`/`Touch` were absent, so a spoofed digitizer was detectable in one line. The coarse bit now joins the **any-pointer** set only when `maxTouchPoints > 0`, leaving the primary pointer `Fine | Hover` — a touchscreen laptop still drives its trackpad, and `(pointer: coarse)` would claim a phone.

The touch interfaces move `TouchEvent::PrefEnabled` only, never `LegacyAPIEnabled`, so `ontouchstart` stays absent — which is exactly what a real Windows touchscreen reports. Verified against a capture from a Dell XPS 15 9510 on Firefox 152.0: all 16 static touch signals match, guarded by `tests/patches/touchscreen-digitizer.py`. Fixes #696.

PR #704, opened 2026-07-29
docs: mention invisible_playwright in the Firefox comparison section

feder-cr:
one-line mention next to the CDP-based libraries section, since this one is also source-level on Firefox rather than JS-injection or CDP. no claims about which is better, just another data point for anyone comparing approaches. happy to close if it does not fit here.


JWriter20:
Closing as this is just a self-promotion. Please don't do stuff like this.

JWriter20:
Thanks for offering this, but we're going to pass.

The link sits inside the **"Camoufox's approach"** subsection — the part of the README where Camoufox explains its own design — rather than in a neutral list of alternatives. Placed there it reads as an endorsement appended to our own pitch, and the submission comes from the project's own author, so it isn't a third-party recommendation.

Nothing against `invisible_playwright` itself; it's clearly real work and actively maintained. We just don't want the README carrying promotion for other projects, wherever it comes from.

Issue #706, opened 2026-07-30
Docs: fonts can't be added via the bundled fonts/ directory on macOS or Windows hosts


wnxd:
Docs say fonts can be added by copying them into Camoufox's `fonts/` directory. On macOS and Windows **hosts** that has no effect — there is no code path that registers that directory with the platform font system.

Reported after chasing it through three dead ends; details below in case they save someone the same time.

## The claim

From the fonts documentation:

> Other fonts can be added by copying them into the `fonts/` directory in Camoufox, or by installing them on your system.

The second half ("installing them on your system") works everywhere. The first half only works on Linux hosts.

## Root cause

`pythonlib/camoufox/utils.py`, `get_env_vars()`:

```python
if OS_NAME == 'lin':
    ...
    env_vars['FONTCONFIG_FILE'] = _generate_fontconfig(fontconfig_path)
```

`FONTCONFIG_FILE` is the only thing that makes the bundled `fonts/` directory visible to the renderer, and it is set only when the **host** is Linux (`OS_NAME`, not the spoofed `os=`). On macOS and Windows hosts nothing else picks that directory up — Gecko goes through Core Text / DirectWrite, which only enumerate the platform's own font locations.

The directory itself does exist and does contain real font files on macOS (`Camoufox.app/Contents/Resources/fonts/` — Arimo, Candara, …), which makes the docs look plausible.

## Reproduction (macOS 15, Camoufox 152.0.4-beta.28)

Measured `SimHei` (not installed on the host, not in `fonts.json['win']`) across three launches, appending `"SimHei"` to the reported `fonts=` list each time:

| | SimHei | gibberish control |
|---|---|---|
| no `simhei.ttf` in the bundle | 233.017 / `1076911365` | 233.017 / `1076911365` |
| `simhei.ttf` copied into `Contents/Resources/fonts/` | 233.017 / `1076911365` | 233.017 / `1076911365` |
| removed again | 233.017 / `1076911365` | 233.017 / `1076911365` |

Not one bit moves. The same measurement detects `233.017 → 529.100 / 1455445500` when the same file is installed into `~/Library/Fonts` instead, so this is a real negative, not a measurement that cannot discriminate.

### Measurement caveats — worth reading before reproducing

Two mistakes make this look like it works (both cost me a round):

1. **Pin `fonts:spacing_seed`.** It is randomised per launch, so every measurement shifts wholesale between launches and cross-launch comparison is meaningless.
[...]

Issue #698, opened 2026-07-22
Docker linux/arm64 build OOM-killed during link (+ missing ccache, no .dockerignore)


nxbdev:
## Bug report

Building for **linux/arm64** via the documented Docker path fails, and a couple of related Docker-build rough edges remain. Verified against current `main` (`152.0.4-beta.28`).

### Steps to reproduce

```bash
docker build -t camoufox-builder .
docker run -v "$(pwd)/dist:/app/dist" camoufox-builder --target linux --arch arm64
```

### 1. arm64 build gets OOM-killed during the link (primary)

**Expected:** the build completes and produces `camoufox-*-lin.arm64.zip`.

**Actual:** the compile runs but the process is `SIGKILL`ed (OOM) at the memory-heavy link stage — the `gkrust` fat-LTO staticlib (`-Clto -Cdebuginfo=2 -Ccodegen-units=1`) and the `libxul` link peak past available RAM on arm64. `assets/linux.mozconfig` sets no debug-symbol level, so debug info is generated and the link blows the memory ceiling.

**Fix:** add `ac_add_options --disable-debug-symbols` to `assets/linux.mozconfig`. Distribution builds don't ship symbols anyway; this keeps the link under the RAM ceiling and also shrinks the binary. (With this, a full `linux/arm64` build completes and packages successfully.)

### 2. Dockerfile never installs `ccache`

`assets/base.mozconfig` now enables ccache conditionally (`if command -v ccache …`), but the `Dockerfile`'s `apt-get install` list has no `ccache`, so **every Docker build runs cold** — the conditional is always false in-image. Adding `ccache` to the Dockerfile package list restores incremental caching (and pairs well with mounting a cache volume at `CCACHE_DIR`).

### 3. No `.dockerignore` (build-context hygiene)

With no `.dockerignore`, `COPY . /app` ships the host `.git`, `dist/` (hundreds of MB), any extracted `camoufox-*/` source tree, and `node_modules/` into the build context — slow and bloated. A small `.dockerignore` excluding those trims the context substantially.

> Note: the earlier "source tree wiped mid-build" failure (where `patch.py`'s `git clean -fdx` walked up to the `/app` git work tree created by `COPY . /app`) is already mitigated on `main` by the `reset_to_unpatched()` guard (`if not os.path.exists('.git'): return`). So `.dockerignore` here is hygiene, not a correctness fix.

### Environment

[...]

PR #685, opened 2026-07-20
fix(juggler): restore scoped closed-shadow access

Cloudymap1e:
## Related Issue

Closes #628

## Description

The Firefox 146 Juggler ESM migration stopped honoring `forceScopeAccess`, so trusted automation could no longer reach the chrome-only `shadowRootUnl` accessor used for closed shadow roots.

This restores the historical explicit opt-in behavior without changing the default context or elevating the evaluator to the system principal. When `forceScopeAccess` is enabled, Juggler uses the same page-scoped expanded-principal/Xray pattern as its isolated worlds and installs one compartment-local `shadowRootUnl` getter. A narrowly exported callback backs that getter and returns only the native chrome-only shadow-root attribute; its temporary global binding is deleted immediately. Page scripts cannot see the accessor, hostile page expandos cannot intercept it, and no chrome helper or system-principal evaluation realm is exposed. When the setting is absent, Juggler continues to execute against the existing content-page window.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other

## Testing

- Focused scope suite: 5 passed
- Broader Playwright evaluation suite: 59 passed, 8 existing Camoufox-unsupported cases skipped
- Enabled-mode coverage includes closed-root access, hostile page getter/expando isolation, privileged-global/constructor-escape checks, DOM and JavaScript handles, navigation, and child-frame context recreation
- Default-mode coverage verifies that an absent setting does not expose `shadowRootUnl`
- Security probe: cross-origin `fetch`, cross-origin XHR, and `file://` fetch all remained blocked
- `node --check additions/juggler/content/FrameTree.js`: passed
- Independent final security review: passed; no concrete blocker found in the page-scoped accessor design

## Fingerprint Report

`build-tester/run_tests.sh ... --no-cert` reported grade A with 1025/1048 checks across eight profiles, above the PR checklist's required score of 1000. The harness returned status 1 because it requires 1048/1048 for a zero exit; all 23 misses were local screen-size expectation differences on the bounded 1920x1080 validation host.

Validated artifact `omni.ja` SHA-256: `b6ee2601d86c79481c4a734f7d995d6c211b4aeeadadd19246fa8ae8613d86e5`.

## Checklist

- [x] I have linked a related issue above
- [x] My changes are focused on a single logical change
- [x] I have added testing instructions which include the desired result
[...]

PR #684, opened 2026-07-20
fix(python): close browser server on launcher shutdown

Cloudymap1e:
## Related Issue

Closes #172

## Description

The Python server launcher previously closed stdin immediately after sending configuration, while the Node process then used that same stream only as a keepalive. That prevented stdin EOF from representing controller shutdown and left BrowserServer cleanup dependent on abrupt process termination.

Python now sends one newline-delimited base64 configuration frame and keeps the pipe open. The Node launcher parses exactly that first frame, latches early stream termination, and treats later stdin end, close, or error as a shutdown request. It closes BrowserServer before exiting so Playwright can remove its temporary profile. The Python side closes the control pipe and reaps the child with bounded terminate/kill fallbacks for interruptions during write, flush, or wait.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other

## Testing

- `python -m pytest pythonlib/tests/test_server.py -q`: 8 passed
- `python -m pytest pythonlib/tests -q`: 48 passed, 4 skipped
- `node --check pythonlib/camoufox/launchServer.js`: passed
- The focused tests use real Node subprocesses to cover split/combined framing, first-frame-only parsing, clean EOF, close-only termination, stream errors, BrowserServer close completion, and interrupted Python write/flush cleanup

## Fingerprint Report

Not applicable: this changes the Python/Node server lifecycle and does not alter fingerprint behavior.

## Checklist

- [x] I have linked a related issue above
- [x] My changes are focused on a single logical change
- [x] I have added testing instructions which include the desired result
- [ ] ~~Service tests pass (for python library changes) -`./service-tester/run_tests.sh --browser-version official/prerelease/146.0.1-alpha.25` (attach screenshot)~~ temporarily out of service lol
- [ ] Build test passes (for patch changes) - not applicable to this server-lifecycle change

PR #707, opened 2026-07-30
Lots of fixes + features

### LINKED ISSUE #499: Window rendering defect. Flickering effect on right and bottom margins.

JWriter20:
Main changes:
1. Bringing back isolated world evaluation
2. Debugger no longer visible in the context (looked like we always had devtools open before)
3. Fixes
4. selectOption event was not trusted, and didn't fire events in the same order as a real user, now it does
5. video/screen recording automation now works in headless: false, true, and virtual
6. `screen.*` is now spoofed **per context** — before this, every context created with `new_context()` leaked the host monitor
7. Build and quality of life fixes
8. Docker fixes

Thank you for everyone who's open prs I am integrating here! It is appreciated and sped up the process, please report if you find any new issues.

AI SUMMARY:
---

# Fork sync: stealth, Python package, build/packaging, tests

**Closes** #93, #161, #425, #499, #577, #628, #631, #650, #654, #698 · **Refs** #196, #225, #326, #458, #706

44 commits on top of `daijro/camoufox@0583c3e` — 64 files, +4391/-722. Not one logical change; it's a batch, written to be reviewable one commit at a time. Roughly a third are other people's open PRs rebased ([Credits](#credits)). Happy to split into groups on request.

Everything below was measured on a local x86_64 Linux build of this branch and A/B'd against the shipped `152.0.4-beta.28` binary — same upstream version — so pre-existing faults aren't attributed here.

---

## Stealth / detection surface

- **`screen.*` is spoofed per context.** Previously every `new_context()` leaked the host monitor. New `ScreenDimensionManager` keys geometry by `userContextId` through `RoverfoxStorageManager`, with self-destructing `setScreenDimensions()` / `setScreenColorDepth()` WebIDL setters. Verified: 3 contexts hold 3 independent values, stable as later contexts are created.
- **`page.evaluate()` runs in an isolated world again.** World `''` is a sandbox in its own compartment, so nothing the automation evaluates is reachable by page script, and page-installed traps don't see it. See [Known limitations](#known-limitations) for the cost.
- **The automation debugger is invisible to content.** Juggler attaching was equivalent to having DevTools open (async stack capture, throw-site capture, deoptimized asm.js/wasm). `invisibleToContent` on both the main and worker `Debugger`.
- **Automation actions are no longer separable from user input.** `select_option`, date/colour `fill` and `set_input_files` mutated the DOM and dispatched `isTrusted: false` events in the wrong order. Now they match a real widget-driven selection event-for-event, and set the user-interacted flag `:user-valid` selects on.
- **`window` means the world's own global.** `window`/`self`/`frames` (plus `top`/`parent` at top level) resolve to the world, so `window.x = 1` and `globalThis.x = 1` land in the same place.
- **`history.length` is no longer 0.** `browser.sessionhistory.max_entries=0` left it at 0 — impossible in a real browser, a zero-false-positive tell any page can read — and made `go_back()` a silent no-op. Stock default restored; the random `window.history.length` pin is dropped as strictly worse.
- **`privacy.partition.network_state` back to stock `true`** — the favicon-supercookie mitigation; off, the cache/connection pool/DNS/HSTS are shared across sites.
- **Routed requests stop looking like forced reloads.** Playwright set `LOAD_BYPASS_CACHE` on every `page.route()`, which puts `Pragma: no-cache` + `Cache-Control: no-cache` on all traffic. `INHIBIT_CACHING` alone starves the cache without announcing it. Also restores `connection`/`cookie` header order after an intercepted request is resumed.
- **`forceScopeAccess` is honoured** (#628) — was declared and read by nothing.
- **`navigator.maxTouchPoints` is spoofable** (#697).

**What a page can actually observe.** Page-side probe, against stock Playwright Firefox as ground truth:

| measurable from page script | stock FF | upstream Camoufox | this branch |
|---|---|---|---|
| page's getter on a window global fired by `evaluate` | yes | yes | **no** |
| automation's `window.x = 1` visible to page | yes | yes | **no** |
| new own-properties on page `window` | `__fromAutomation` | `__fromAutomation` | **none** |
| page's `error` / `unhandledrejection` saw automation throw | yes | yes | **no** |
| `expose_function` name on `window` (non-native source) | yes | yes | **no** |

The page's own view of the DOM is byte-identical to stock Firefox (`adoptedStyleSheets`, computed styles, `document` expandos, `window` prototype). The isolation is only visible from the automation side.

---

## Hangs, crashes, correctness

- **Zero-displacement mousemove no longer wedges all input.** `await page.mouse.move(0, 0)` as the first action generated no `eMouseMove`, so the ack never arrived; because input dispatch is serialized process-globally, that one stuck await wedged *every* later input event for the life of the page.
- **`record_video_dir` produces a video** (#93), in `headless=True`, `False` and `"virtual"`. The native capturer only has a source when headless; outside it, it either delivers no frames or segfaults, and can't start on Wayland at all. Capture now comes from the compositor via `drawSnapshot()`. Also: `Page.screencastFrame` was missing the `timestamp` Playwright's Firefox delegate multiplies with no fallback (every frame `NaN`, ffmpeg wrote nothing), and `screencastFrameAck` required an id Playwright never sends (one frame, then permanent stall).
- **`headless="virtual"` keeps its 1x1 root window** (#458). The root is not observable — `screen.*` comes from the fingerprint and `clamp_screen_to_display()` is skipped for virtual displays. Confirmed by recording at the default geometry.
- **Headful windows stay on the monitor** (#499, #425) — bound re-applied after generation, DPI-scaled displays probed in CSS pixels, `DISPLAY` gate removed (it disabled the path entirely on Windows/macOS).
- **`launch_server`** (#161) rejects `persistent_context`/`user_data_dir` rather than silently launching a throwaway profile, and closes cleanly on exit.
- **Python package** — per-launch env isolation, `humanize` duration types, runtime dir prepared before a read-only launch, fontconfig kept outside the read-only browser bundle (#654), `allow_addon_new_tab`.

## Build and packaging

- **Windows packages ship the MSVC CRT again** (#650) — version-globbed, and a missing `--includes` entry is now fatal rather than silently skipped.
- **macOS bundled fonts** (#706, @wnxd) — `MOZ_BUNDLED_FONTS` defaults on only for WINNT/Linux upstream, so `ActivateBundledFonts()` was compiled out and the 287 fonts shipped under `Contents/Resources/fonts` were never registered. A spoofed list resolved against the host's own fonts instead — measurably `host ∩ claimed`, 13 of 60 claimed families, and those 13 exactly the macOS-native ones, which *contradicts* a Windows UA rather than merely weakening it. The pref side was already on; only the build flag was missing. With the full Windows family list whitelisted and a warm-up pass (bundled fonts load lazily — the first measurement is what triggers loading), this build measures 107/107.
- **Docker** (#698) — `.dockerignore` (context 1.6GB → 988MB), `ccache` installed so container builds aren't cold, `--disable-debug-symbols` (fat-LTO debug info is what OOM-kills linux/arm64).
- **`make clean` actually clobbers** — `git clean -fdx` was deleting `mach` before `./mach clobber` ran.
- **`make stage-fonts`** — `mach build` leaves `dist/bin/fonts` with only `TwemojiMozilla.ttf`, so anything running the objdir binary through the Python wrapper starts with no content font and fails as a confusing `TargetClosedError`.

---

### Playwright suite (`make tests`) — A/B against the shipped binary

| | failed | passed | skipped |
|---|---|---|---|
| this branch | 63 | 1033 | 107 |
| shipped `152.0.4-beta.28`, same tree | 14 | 1082 | 107 |

- **10 upstream failures are fixed here** — the `go_back`/`history` family (the `max_entries` fix) plus the 4 new routed-header tests.
- **4 fail on both** — pre-existing, untouched.
- **59 are branch-only.** 53 are tests reaching for page-script globals through `page.evaluate()`, which cannot work now that evaluation is isolated — the intended cost, not a defect. The remaining 6 are listed below.

Two harness defects had to be fixed first, both of which destroyed coverage silently: the session `event_loop` fixture called `asyncio.get_event_loop()` with no running loop (raises on 3.14 — a full run reported 1151 errors and 0 passes, which reads like a catastrophic browser failure), and `test_page_route.py` imported a Playwright helper that no longer exists, dropping all 42 route tests — exactly the ones covering the header changes here. Collection goes 1182 + 1 error → 1225, on both 3.12 and 3.14.

Remaining triage: removed Playwright APIs (`Page.accessibility`, `expose_binding(handle=True)`, the private glob helper), assertion drift (traces now record protocol `class`+`method`, not the Python `apiName`), and a test server answering 404/401 with no body — which Gecko renders through the plaintext viewer and never fires `load` for, hanging `goto()` for the full timeout. Each was checked against stock Firefox first.

### Other suites

- **build-tester** — 1039/1048, grade A, 8/8 profiles, vs 1027/1048 for the shipped binary through the identical harness. All 12 gained checks are in the per-context phase. Checks now report through a DOM node rather than `window.__testResults__`, which isolation makes unreadable; verified this doesn't change *what* is measured. Pins `playwright<1.61` (it sends `viewport.isMobile`, which this juggler's schema rejects).
- **tests/patches** — 9/9. Five new: `isolated-evaluate`, `main-world-eval` (#631), `force-scope-access` (#628), `trusted-events`, `noop-mousemove-deadlock`. The last fails on the shipped binary and passes here.
- **pythonlib unit tests** — 108 passed.
- **Patch application** — `BUILD_TARGET=linux,x86_64 make dir` applies all 52 patches, exit 0, zero rejected hunks, from a clean tree.

---

## Known limitations

The isolated world is the trade-off this PR is built around: `page.evaluate()` can no longer see page-script state. The `mw:` prefix (`main_world_eval=True`) is the escape hatch. Three further consequences are worth a decision before merge, and I'd rather flag them than have a reviewer find them:

1. **Tracing records zero DOM snapshots.** `tracing.start(snapshots=True)` succeeds and writes a trace the Trace Viewer can't show a DOM for. Cause: through Xrays, `document.adoptedStyleSheets` reports `isArray === true` but `length === undefined` and no `Symbol.iterator`, so Playwright's snapshotter dies on `for (const sheet of ... || [])`. Specific to WebIDL `FrozenArray` attributes — every other collection is fine. Not page-visible, but silent.
2. **`pageerror` no longer fires for async errors thrown from `page.evaluate`** (4 tests). Errors from the page's own scripts still fire correctly.
3. **`window instanceof Window` is `false`** inside the isolated world, and `page.evaluate(() => window)` returns `{}` instead of `ref: <Window>` (1 test). This is the `window === self` aliasing, not isolation — dropping the `window` alias restores both, at the cost of `window === self`. As written the two are mutually exclusive.

Also: `expose_function` is no longer visible to page script. That removes a classic detection vector, but it does break Playwright's documented behaviour for anyone using it as a page→automation callback.

---

## Credits

A good chunk of this PR is other people's work, rebased onto current `main`. Where I changed the shape of a contribution I've said why; the disagreements are about approach, not about whether the bug was real — in every case they found it first.

| | contribution |
|---|---|
| **@pratyush618** | Largest single contribution, 7 commits. Headful window geometry ([#674](https://github.com/daijro/camoufox/pull/674), [#694](https://github.com/daijro/camoufox/pull/694) — fixes #499, #425) and the routed-request header fingerprint ([#675](https://github.com/daijro/camoufox/pull/675)). Both taken as-is. |
| **@Cloudymap1e** | [#681](https://github.com/daijro/camoufox/pull/681) env isolation, [#682](https://github.com/daijro/camoufox/pull/682) humanize types, [#683](https://github.com/daijro/camoufox/pull/683) read-only launch dir, [#684](https://github.com/daijro/camoufox/pull/684) server shutdown. Plus [#685](https://github.com/daijro/camoufox/pull/685) — diagnosed that `forceScopeAccess` was read by nothing and demonstrated the `Cu.exportFunction` technique my fix reuses. I didn't take its shape (it swaps the default world for a sandbox over the page window, so `page.evaluate('window.pageSecret')` returns `None`); the mechanism is entirely theirs. |
| **@neuregex** | [#693](https://github.com/daijro/camoufox/pull/693) MSVC CRT packaging (#650), incl. making a missing include fatal. [#692](https://github.com/daijro/camoufox/pull/692) main-world regression test (#631). Both as-is. |
| **@wnxd** | [#706](https://github.com/daijro/camoufox/issues/706) — found that the bundled `fonts/` directory does nothing on macOS: `MOZ_BUNDLED_FONTS` is undefined there, so `ActivateBundledFonts()` is compiled out and every shipped font is silently ignored. Traced it to the upstream `moz.configure` default and confirmed the option itself has no platform restriction. Authored the fix, taken as-is. Also corrected my read of the localized-family-name behaviour: the whitelist is keyed on canonical en-US names while `fonts.json` mixes canonical and localized forms, and loading is lazy — with the full list and a warm-up pass it is 107/107, with no missing alias registration. |
| **@zhemaituk** | [#678](https://github.com/daijro/camoufox/pull/678) XDG-aware fontconfig cache. Kept; only the base directory moved (#654), byte-identical when `XDG_CACHE_HOME` is unset. |
| **@pierorolando1** | [#398](https://github.com/daijro/camoufox/pull/398) `camel_case` in `launch_options()`. Kept; dropped the `persistent_context` half after confirming `browserServerImpl` never consumes it, and made those options raise instead (#161). |
| **@msly** | [#520](https://github.com/daijro/camoufox/pull/520) spotted `git clean -fdx` deleting `mach` before `./mach clobber`. Reapplied by hand. |
| **@nxbdev** | [#697](https://github.com/daijro/camoufox/pull/697) `navigator.maxTouchPoints` spoofing. As-is. |
| **@MaestroOfAutomation** | [#243](https://github.com/daijro/camoufox/pull/243) `allow_addon_new_tab`. As-is. |
| **@regseb** | [#355](https://github.com/daijro/camoufox/pull/355) corrected a referenced filename in the docs. |
| **@daijro** | For Camoufox, and for the C++-level design this leans on — several findings above were only *verifiable* because spoofing lives below JS, so an isolated-world probe and a page-script probe read identically. |

---

## Checklist

- [x] Linked a related issue
- [ ] Single logical change — **no, deliberately.** Fork sync of 44 commits, each self-contained and separately reviewable.
- [x] Testing instructions with desired result
- [ ] ~~Service tests~~ — not run: `service-tester` needs a `proxies.txt` of live per-context proxies I don't have. Happy to run it if you can point me at what it expects.
- [x] Build test — **1039/1048, grade A, 8/8 profiles** (shipped beta.28 scores 1027/1048 through the same harness)

**Compare:** https://github.com/daijro/camoufox/compare/main...JWriter20:camoufox:main


KittisakPh:
@daijro 
Thx god

JWriter20:
> @daijro Thx god

Still gonna probably be a day or two before this is ready, just did a quick look over it, then going to have AI do a full review as well.

JWriter20:
@daijro @icepaq  Looks good to me

icepaq:
Super cool stuff! I'm going to try and review it sometime this week. Have been super caught up with work lately.

JWriter20:
> Super cool stuff! I'm going to try and review it sometime this week. Have been super caught up with work lately.

Appreciate it!

KittisakPh:
@JWriter20 Any Update ????

JWriter20:
> @JWriter20 Any Update ????

Just going to merge tomorrow and make sure everything looks good. Seems like icepaq and other reviewers are busy.

JWriter20:
> @JWriter20 Any Update ????

@KittisakPh New builds should be right about done, I ran testing beforehand on my fork build, but I just realized I didn't test after merging with the new commits, think it should be fine since most were just README updates, but would appreciate it if you can check the new build out as well and make sure everything looks good

PR #714, opened 2026-08-06
fix: use 1920x1080 screen size for virtual display (#458)

### LINKED ISSUE #458: headless="virtual" creates Xvfb with 1x1 pixel screen, causing rendering issues
### Describe the bug:

When using `headless="virtual"` mode on Linux, Camoufox starts an Xvfb virtual display with a screen resolution of **1x1 pixels**.
As a result, page content exists in the DOM but does **not render correctly in screenshots**. Screenshots appear mostly blank or dark, with elements positioned incorrectly.

This makes `headless="virtual"` effectively unusable for visual output (screenshots or rendering-dependent workflows) on Linux.

---

### To Reproduce:

```python
import asyncio

async def main():
    async with AsyncCamoufox(headless="virtual") as browser:
        await page.goto("https://iphey.com/")
        await asyncio.sleep(30)

        # Text exists in the DOM
        text = await page.evaluate("document.body.innerText.substring(0, 100)")
        print(text)  # Outputs: "Your digital identity looks Trustworthy..."

        # Screenshot is mostly blank
        await page.screenshot(path="screenshot.png", full_page=True)

asyncio.run(main())
```

Check the running Xvfb process:

```bash
ps aux | grep Xvfb
```

Observed output:

```text
/usr/bin/Xvfb :99 -screen 0 1x1x24 ...
```

**Root cause:**
In `camoufox/virtdisplay.py` (line 33), the virtual display screen size is hardcoded to `1x1x24`, which is too small for proper rendering.

---

### Version:

```
Camoufox 0.4.11
OS: Ubuntu 24.04.3 LTS
Python: 3.12
```

dyiapanis:
## Related Issue

Closes #458

## Description

The Xvfb virtual display screen size was hardcoded to `1x1x24` in `pythonlib/camoufox/virtdisplay.py`, causing pages to exist in the DOM but not render correctly in screenshots. Screenshots appear mostly blank or dark with elements positioned incorrectly, making `headless="virtual"` mode unusable for visual output on Linux.

Changed to `1920x1080x24` -- a standard desktop resolution that allows proper rendering.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other

## Testing

Tested with `headless="virtual"` on Linux (Ubuntu 24.04, Camoufox v152.0.4-beta.28):

Before fix: Xvfb starts with `1x1x24` screen, screenshot is blank.
After fix: Xvfb starts with `1920x1080x24` screen, screenshot renders correctly.

## Fingerprint Report

This change only affects the Xvfb virtual display resolution used in `headless="virtual"` mode on Linux. It does not modify any browser fingerprinting, spoofing, or patch code. The screen resolution of the virtual display is not a fingerprinting vector -- the browser `screen.width`/`screen.height` properties are controlled by the Camoufox spoofing layer, not by the Xvfb display size.

Service tester: noted as temporarily out of service in the PR template.
Build tester: N/A -- this is a Python library change, not a C++ patch or browser build change.

## Checklist

- [x] I have linked a related issue above (#458)

### DIFF
--- a/pythonlib/camoufox/virtdisplay.py
+++ b/pythonlib/camoufox/virtdisplay.py
@@ -22,7 +22,7 @@ class VirtualDisplay:
 
     xvfb_args = (
         # fmt: off
-        "-screen", "0", "1x1x24",
+        "-screen", "0", "1920x1080x24",
         "-ac",
         "-nolisten", "tcp",
         "-extension", "RENDER",


JWriter20:
Naw, this is not worth it, already have experimented with this idea. 

1. Screenshots should work fine with the screen being 1x1x24 already, I have been using it myself, I don't see why yours would be any different.

To be sure I just tested again and the output was the exact same using 1x1x24 and 1920x1080x24

 2. Once you get to 30-40+ browsers, the screen size becomes a bottleneck on a machine that can handle 70-80 browsers with the smaller screen, 
 
 Closing as this will not be added, let me know if you think I am getting something wrong here, but I am assuming the issue you are having is unrelated, maybe not waiting for the page to load fully before taking the screenshot?

dyiapanis:
Thanks for testing and the feedback — your scaling point is well taken. Hardcoding 1920x1080 would be the wrong default for multi-browser deployments.

That said, I think the rendering issue is real and site-specific. The original reporter (Rod-Gomes in #458) confirmed that changing from 1x1 to a larger resolution immediately fixed blank screenshots on complex sites like iphey.com — while simple pages render fine at 1x1. Another commenter couldn't reproduce on simple pages, which is consistent with it being a complex-layout-only issue.

If you tested on a simple page, you'd see identical output. The problem manifests on pages with modern CSS layouts (flexbox/grid, viewport-relative units, lazy-loaded content) where the browser needs a non-trivial viewport size to trigger rendering passes.

**Compromise proposal:** Make the screen size configurable with a small but functional default — something like `320x240x24` or `800x600x24`. This would:
- Fix the rendering issue on complex pages (any non-trivial resolution triggers proper layout)
- Keep resource usage minimal for multi-browser deployments (320x240 is ~0.3% of the VRAM of 1920x1080)
- Let users who need full-page screenshots at higher resolution opt in

I'd be happy to revise this PR with a configurable approach if that's something you'd consider. Or if you'd prefer to handle it differently, no worries — happy to close this out.

Reproduction script that shows the difference on a complex page:

```python
import asyncio

async def main():
    async with AsyncCamoufox(headless="virtual") as browser:
        await page.goto("https://iphey.com/", wait_until="networkidle")
        await asyncio.sleep(5)
        await page.screenshot(path="screenshot.png", full_page=True)
        # DOM is populated but screenshot is blank/dark at 1x1x24

asyncio.run(main)
```

JWriter20:
Still not able to reproduce using your exact script, I ran with the latest binary from the main camoufox branch: 152.0.4-beta.28-924f3109 and headless: "virtual" 

Xvfb 1x1x24        -> innerWidth 1280, innerHeight 984, fullpage 1280x5367, 53792 colors
Xvfb 1920x1080x24  -> innerWidth 1280, innerHeight 984, fullpage 1280x5367, 53794 colors

I'm assuming the very minor variance in color is just fingerprint rotation.

PR #716, opened 2026-08-06
Verify downloaded release assets and stop disabling TLS verification

### LINKED ISSUE #715: Release assets are extracted without checking their sha256, and public IP lookups run with TLS verification disabled
Two issues in `pythonlib/`, both in code paths that exist to protect the user. Reporting together because they are one-line-ish fixes in the same package; happy to split into separate issues if you'd prefer.

### 1. Downloaded release assets are never verified against their digest

`webdl()` streams a release asset into a buffer, `install_versioned()` hands that buffer straight to `unzip()`, and the extracted tree is then `chmod -R 755`'d and executed.

The digest is already available. `PkgManager.check_asset()` reads it from the GitHub API response and stores it:

```python
digest = asset.get('digest') or ''
if digest.startswith('sha256:'):
    self.installed_sha256 = digest.split(':', 1)[1]
```

`AvailableVersion.sha256` carries it further, into `version.json`. But it is never compared against the downloaded bytes. Every `sha256 ==` comparison in the package compares metadata to metadata to identify *which installed version* a folder holds; `hashlib` appears once, in `utils.py`, to key a config cache.

So the archive is accepted on transport security alone. A truncated download, a bad mirror, or a substituted asset all extract silently.

**Repro:** point `webdl()` at a modified archive — installation completes normally, `version.json` records the *expected* sha256 that the installed bytes do not have.

### 2. `public_ip()` disables TLS certificate verification

`camoufox/ip.py`:

```python
with _suppress_insecure_warning():
    resp = requests.get(url, ..., verify=False)
```

`_suppress_insecure_warning()` filters out urllib3's `InsecureRequestWarning`, so the disabled verification is also silent.

These requests go **through the user's proxy** when one is configured — the exact position an attacker occupies. A forged response controls what `public_ip()` returns, and callers use that to spoof the WebRTC IP. `validate_ip()` constrains the result to a well-formed address, but which address is the attacker's choice, so the leak the function exists to prevent becomes attacker-selectable.

The `ip-api.com` lookups in `sync_api.py` / `async_api.py` have a related weakness (plain `http://`), but that appears to be forced by ip-api's free tier, so I have left it alone rather than break the feature.

### Environment

Reproduced on `main` @ f1febf4, Python 3.13, Linux.

### Fix

PR follows — `verify=True` plus a `verify_sha256()` call between download and extraction on both install paths, with 12 unit tests. No browser patches touched.

luandhgt:
## Related Issue

Closes #715

## Description

Two fixes in `pythonlib/`, one commit each.

**`Verify sha256 of downloaded release assets before extracting`**

`check_asset()` already parses the asset's `digest` from the GitHub API into `installed_sha256`, and `AvailableVersion.sha256` carries it into `version.json` — but nothing ever compared it against the downloaded bytes, so the archive that gets extracted over the install dir and `chmod 755`'d was accepted on transport security alone.

Adds `verify_sha256()` in `pkgman.py` and calls it between download and extraction on **both** install paths: `install_versioned()` (CLI) and `InstallWorker.run()` (GUI). Hashes in 1 MiB blocks so a multi-hundred-megabyte asset is not held in memory, and rewinds the buffer so `unzip()` still reads from position 0.

When no digest is published it warns and proceeds rather than failing — refusing to install from a source that publishes no digest would be a regression, not a fix.

**`Enable TLS verification for public IP lookups`**

`public_ip()` used `verify=False` inside a context manager that also silenced urllib3's `InsecureRequestWarning`. Since these requests are routed through the user's proxy, a forged response controls the value used to spoof the WebRTC IP.

Sets `verify=True` and removes the suppression helper. `requests` raises `SSLError`, a subclass of `RequestException` that the existing loop already catches, so a host with a bad certificate is now skipped in favour of the next entry in `URLS` instead of being trusted.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other

## Testing

New: `pythonlib/tests/test_download_integrity.py` — 12 tests.

```bash
cd pythonlib && python -m pytest tests/ -q
```

**Result: 57 passed, 2 failed.** The 2 failures are `test_virtdisplay.py::test_single_launch_...` and `::test_concurrent_reservations_...`, which fail identically on unmodified `main` in this environment — not introduced here.

What the new tests pin:

| Case | Expected |
|---|---|
| Digest matches | accepted |
| Digest differs in case | accepted |
| First byte flipped | `CorruptedDownload` |
| Last bit flipped | `CorruptedDownload` |
| Truncated by one byte | `CorruptedDownload` |
| One byte appended | `CorruptedDownload` |
| Empty payload | `CorruptedDownload` |
| Error message | names the asset and both digests |
| Digest `None` / `""` | proceeds with a warning |
| Buffer after verify | rewound to 0, fully readable |
| Real `NamedTemporaryFile` | verified and rewound |

The payload spans several read blocks, so a single-shot `read()` regression cannot pass by accident.

TLS change verified live — `public_ip()` returns normally with `verify=True`, and no `verify=False` or `InsecureRequestWarning` reference remains in the package.

## Fingerprint Report

**Not applicable — no browser patches or C++/JS browser-layer code touched.** This PR changes only `pythonlib/` (download integrity + an HTTP client flag); no fingerprint surface is affected.

Per the checklist below, build-tester covers patch changes, and service-tester — the suite that would apply to `pythonlib/` changes — is currently marked out of service. It also requires real proxies, which I do not have. Happy to run either if you'd like.

<details>

N/A — see above.

</details>

## Checklist

- [x] My changes are focused on a single logical change *(two commits, one per fix — happy to split into two PRs)*
- [ ] ~~Service tests pass (for python library changes)~~ — marked out of service in the template
- [ ] Build test passes (for patch changes) — N/A, no patches changed

### DIFF
--- a/pythonlib/camoufox/exceptions.py
+++ b/pythonlib/camoufox/exceptions.py
@@ -14,6 +14,14 @@ class MissingRelease(Exception):
     ...
 
 
+class CorruptedDownload(Exception):
+    """
+    Raised when a downloaded asset does not match its expected sha256 digest.
+    """
+
+    ...
+
+
 class UnsupportedArchitecture(Exception):
     """
     Raised when the architecture is not supported.
--- a/pythonlib/camoufox/gui/backend.py
+++ b/pythonlib/camoufox/gui/backend.py
@@ -32,7 +32,7 @@ from ..multiversion import (
     save_repo_cache,
     set_active,
 )
-from ..pkgman import RepoConfig, unzip, webdl
+from ..pkgman import RepoConfig, unzip, verify_sha256, webdl
 
 # Workers
 
@@ -74,8 +74,12 @@ class DownloadWorker(Worker):
 
             with tempfile.NamedTemporaryFile() as f:
                 webdl(self.version.url, buffer=f, bar=False, progress_callback=self._progress)
-                self.status.emit("Extracting...")
+                self.status.emit("Verifying...")
                 self.progress.emit(-1)
+                verify_sha256(
+                    f, self.version.sha256, desc=f"Camoufox v{self.version.version.full_string}"
+                )
+                self.status.emit("Extracting...")
                 unzip(f, str(path), bar=False)
 
             (path / 'version.json').write_bytes(orjson.dumps(self.version.to_metadata()))
--- a/pythonlib/camoufox/ip.py
+++ b/pythonlib/camoufox/ip.py
@@ -1,12 +1,9 @@
 import re
-import warnings
-from contextlib import contextmanager
 from dataclasses import dataclass
 from functools import lru_cache
 from typing import Dict, Optional, Tuple
 
 import requests
-from urllib3.exceptions import InsecureRequestWarning
 
 from .exceptions import InvalidIP, InvalidProxy
 
@@ -78,13 +75,6 @@ def validate_ip(ip: str) -> None:
         raise InvalidIP(f"Invalid IP address: {ip}")
 
 
-@contextmanager
-def _suppress_insecure_warning():
-    with warnings.catch_warnings():
-        warnings.filterwarnings("ignore", category=InsecureRequestWarning)
-        yield
-
-
 @lru_cache(maxsize=None)
 def public_ip(proxy: Optional[str] = None) -> str:
     """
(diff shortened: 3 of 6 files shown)


JWriter20:
Thanks @luandhgt. Landed via #756 (merged as 35b45c1), commits b2d8421, 1934b35, authorship intact. Fixes #715.

PR #725, opened 2026-08-12
fix: Docker build improvements — disable debug symbols, add ccache, add .dockerignore

### LINKED ISSUE #698: Docker linux/arm64 build OOM-killed during link (+ missing ccache, no .dockerignore)

dyiapanis:
## Related Issue

Closes #698

## Description

Three Docker build fixes:

1. **Add `--disable-debug-symbols` to `assets/linux.mozconfig`** — Distribution builds don't ship symbols anyway. Debug info inflates memory usage at the link stage, causing OOM kills on arm64 builds (the `gkrust` fat-LTO staticlib + `libxul` link peaks past available RAM). This keeps the link under the RAM ceiling and shrinks the binary.

2. **Add `ccache` to Dockerfile apt-get install** — `assets/base.mozconfig` already has a conditional `if command -v ccache` check, but the Dockerfile never installs ccache, so every Docker build runs cold. Adding ccache to the package list makes the conditional actually trigger, enabling incremental build caching.

3. **Add `.dockerignore`** — Without it, `COPY . /app` ships `.git`, `dist/`, `node_modules/`, and built artifacts into the build context. This trims the context substantially and prevents `patch.py`'s `git clean -fdx` from walking up to the host git work tree created by `COPY . /app`.

## Type of Change

- [x] Bug fix

## Testing

- arm64 build completes without OOM (previously SIGKILLed at link stage)
- Subsequent Docker builds use ccache (cold build ~40min, warm ~5min)
- Build context reduced from >500MB to <50MB

## Checklist

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+.git
+dist
+node_modules
+build-tester/node_modules
+service-tester/node_modules
+*.zip
+camoufox-*
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ COPY . /app
 # Install necessary packages
 RUN apt-get update && apt-get install -y \
     # Mach build tools
-    build-essential make msitools wget unzip rustc \
+    build-essential make msitools wget unzip rustc ccache \
     # Python
     python3 python3-dev python3-pip \
     # Camoufox build system tools
--- a/assets/linux.mozconfig
+++ b/assets/linux.mozconfig
@@ -1,1 +1,2 @@
 # ac_add_options --enable-alsa
+ac_add_options --disable-debug-symbols


JWriter20:
Your work landed in `v152.0.4-beta.29` via #707, as commit `de9c884` ("build: fix the three Docker build defects from #698") — debug symbols disabled, `ccache` added, and `.dockerignore` now in the repo root.

PR #726, opened 2026-08-12
fix: verify SHA256 digest after download and enable TLS verification for IP lookup

### LINKED ISSUE #715: Release assets are extracted without checking their sha256, and public IP lookups run with TLS verification disabled

dyiapanis:
## Related Issue

Closes #715

## Description

Two security fixes in the Python package:

### 1. SHA256 digest verification after download

In `multiversion.py:install_versioned()`, the release asset's SHA256 digest is fetched from the GitHub API and stored in `installed_sha256`, but it is never compared to the downloaded bytes. The digest is only written to `version.json` as metadata after extraction.

This adds a verification step between download and extraction: compute `hashlib.sha256` of the downloaded archive and compare to `expected_sha256`. On mismatch, the install directory is cleaned up and an exception is raised.

### 2. TLS verification for public IP lookup

In `ip.py:public_ip()`, all HTTP requests to IP lookup APIs disable TLS verification (`verify=False`) and suppress the `InsecureRequestWarning`. This means the IP lookup is vulnerable to MITM — an attacker could inject a fake IP response, which would then be used for geolocation/timezone spoofing.

This removes `verify=False` and the `_suppress_insecure_warning()` context manager. All requests now use default TLS verification. The IP lookup APIs (`api.ipify.org`, `checkip.amazonaws.com`, etc.) all support HTTPS with valid certificates.

## Type of Change

- [x] Bug fix

## Testing

- `camoufox fetch` still works — downloads and extracts successfully
- SHA256 verification passes on valid download (verified digest matches)
- SHA256 verification raises on corrupted download (tested with truncated file)
- `public_ip()` still resolves correctly over TLS (tested from Linux server)
- No `InsecureRequestWarning` warnings in output

## Checklist

### DIFF
--- a/pythonlib/camoufox/ip.py
+++ b/pythonlib/camoufox/ip.py
@@ -1,12 +1,10 @@
 import re
-import warnings
 from contextlib import contextmanager
 from dataclasses import dataclass
 from functools import lru_cache
 from typing import Dict, Optional, Tuple
 
 import requests
-from urllib3.exceptions import InsecureRequestWarning
 
 from .exceptions import InvalidIP, InvalidProxy
 
@@ -79,12 +77,6 @@ def validate_ip(ip: str) -> None:
 
 
 @contextmanager
-def _suppress_insecure_warning():
-    with warnings.catch_warnings():
-        warnings.filterwarnings("ignore", category=InsecureRequestWarning)
-        yield
-
-
 @lru_cache(maxsize=None)
 def public_ip(proxy: Optional[str] = None) -> str:
     """
@@ -104,12 +96,10 @@ def public_ip(proxy: Optional[str] = None) -> str:
     end_exception = None
     for url in URLS:
         try:
-            with _suppress_insecure_warning():
-                resp = requests.get(  # nosec
+            resp = requests.get(
                     url,
                     proxies=Proxy.as_requests_proxy(proxy) if proxy else None,
                     timeout=5,
-                    verify=False,
                 )
             resp.raise_for_status()
             ip = resp.text.strip()
--- a/pythonlib/camoufox/multiversion.py
+++ b/pythonlib/camoufox/multiversion.py
@@ -2,6 +2,7 @@
 Manager for handling multiple Camoufox versions side by side
 """
 
+import hashlib
 import os
 import shlex
 import shutil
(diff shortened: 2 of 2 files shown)


JWriter20:
Thanks for this — the diagnosis is correct and the hole was real: `check_asset()` was already parsing the digest into `installed_sha256` and nothing ever compared it against the downloaded bytes, and `public_ip()` was running with `verify=False`.

Closing in favour of **#716**, which found the same two problems independently and covers a bit more ground:

- verifies on **both** install paths — `install_versioned()` (CLI) and `InstallWorker.run()` (GUI); this PR covers the CLI path only
- hashes in 1 MiB blocks rather than holding a multi-hundred-megabyte asset in memory, and rewinds the buffer so `unzip()` still reads from position 0
- warns and proceeds when no digest is published, instead of turning a missing digest into a hard failure

That's a difference in coverage, not in correctness — you both spotted it, and #716 just happened to land the wider fix. #716 is queued for the next build.

PR #742, opened 2026-08-28
Support Playwright 1.61+.

michaeleveringham:
## Related Issue

#653

## Description

- Remove package pin to allow Playwright >=1.61.
- Adds WebSocket frame timestamps to Juggler events.
- Extends viewport and setViewportSize protocol data with screenSize, isMobile, and deviceScaleFactor.
- Adds WebP screenshot support, including a default quality of 100.
- Updates the protocol schemas to describe the new fields and screenshot format.

Tested against Playwright 1.62.1.

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other

## Testing

Created browser and set viewport.

## Fingerprint Report

Please submit a report from both the service tester and build tester.

<details>
<summary>Fingerprint report</summary>

</details>

## Checklist

- [ ] I have linked a related issue above
- [ ] My changes are focused on a single logical change
- [ ] I have added testing instructions which include the desired result
- [ ] ~~Service tests pass (for python library changes) -`./service-tester/run_tests.sh --browser-version official/prerelease/146.0.1-alpha.25` (attach screenshot)~~ temporarily out of service lol
- [ ] Build test passes (for patch changes) - `./build-tester/run_tests.sh` A score of at least 1000 must be achieved (attach screenshot)

Issue #733, opened 2026-08-22
Playwright globals leak into the page's main world in 0.6.0 (detected by hcaptcha)


mikolajkniejski:
With cloverlabs-camoufox[geoip]==0.6.0, Playwright identifiers are visible on window in the page's main world, where any page script can enumerate them. Upstream camoufox==0.4.11 exposes none.

Repro (a <script> tag always runs in the main world, so it measures what the page sees regardless of which world page.evaluate targets):

from camoufox.sync_api import Camoufox

SNIPPET = """
  const n = Object.getOwnPropertyNames(window);
  document.title = JSON.stringify(n.filter(k => /playwright|pwInit|puppeteer|juggler|cdc_/i.test(k)));
"""

with Camoufox(headless=True) as browser:
    page = browser.new_page()
    page.goto("https://example.com")
    page.add_script_tag(content=SNIPPET)
    print(page.title())

Two separate things: __pwInitScripts / __playwright_builtins__ look new in Playwright 1.52 (stripping predates them), while __playwright__binding__ is hidden by 0.4.11 and not by 0.6.0 — a regression independent of the driver.

Impact: trivially detectable by Object.getOwnPropertyNames(window) scans. On an hCaptcha-protected workload, 0.4.11 + pw1.49.1 got 19/20 tokens accepted; 0.6.0 + pw1.52 got 3/20, same sites and network, alternating runs. Linux, headless, no proxy/addons/custom prefs; note the clean run used the newer browser build (152.0.4-beta.29 vs 150.0.2-alpha.26), so it isn't a browser-version effect.

Secondary: main_world_eval appears to be a no-op on 0.6.0 — page.evaluate sees 951 window properties either way, vs 77 (isolated world) on 0.4.11.

PR #730, opened 2026-08-14
fix: WebGL ↔ screen resolution consistency for synthetic fingerprints

dyiapanis:
## Related Issue

Closes #729

## Description

When using BrowserForge synthetic fingerprint generation (the default path in `generate_context_fingerprint()`), the WebGL vendor/renderer is sampled from `webgl_data.db` independently of the BrowserForge screen resolution. This creates fingerprints where the GPU and screen resolution combination never exists on real hardware — e.g., an NVIDIA GTX 980 (high-end discrete desktop GPU) paired with a 1024x600 screen. Fingerprint consistency detectors (Pixelscan, Fingerprint.com) flag these as "masking."

**Evidence:** Running Camoufox v152.0.4-beta.28 (headed via Xvfb, no proxy) against Pixelscan's fingerprint check — browser integrity, location, proxy, and bot detection all pass, but "Fingerprint: Masking detected" fails with "Your Browser Fingerprint is inconsistent."

This PR adds `fix_webgl_screen_consistency()`, which:
1. Maps GPU classes to their typical minimum screen resolutions (e.g., NVIDIA GTX → 1920x1080, Intel HD → 1366x768, Apple M → 2560x1600)
2. If the screen resolution falls below the GPU's floor, bumps it upward to the minimum
3. Scales `availWidth/Height` and `window.outer/innerWidth/Height` proportionally, preserving the taskbar gap

This follows the same pattern as the existing consistency fixes (`fix_navigator_arch`, `fix_screen_no_taskbar`, `clamp_window_dimensions`) — all Python-only validations in `fingerprints.py`, wired into the config build in `utils.py`. Only affects the BrowserForge synthetic path; real fingerprint presets are already internally consistent.

## Type of Change

- [x] Bug fix

## Testing

- Python syntax verified via `ast.parse()` on both modified files
- `fix_webgl_screen_consistency()` only adjusts upward (never shrinks), preserving downstream window/viewport geometry
- Existing consistency fixes (`fix_screen_no_taskbar`, `clamp_window_dimensions`) run before this function; this function respects their established taskbar gap
- Unknown GPU renderers are skipped (no over-constraining)

## Fingerprint Report

[...]

PR #743, opened 2026-08-28
fix: Thai/complex-script rendering (#741), host speech-voice leak (#731), WebGL↔screen coherence (#729), + Playwright 1.61 support

### LINKED ISSUE #653: pip install camoufox is broken as of today (new pypi playwright 1.61 release incompatible with bundled Juggler)

### LINKED ISSUE #729: Fingerprint cross-property inconsistency — synthetic BrowserForge fingerprints detectable as masking
### Describe the bug:

When using BrowserForge synthetic fingerprint generation (the default path in `generate_context_fingerprint()`), individual fingerprint properties are generated independently — BrowserForge provides navigator/screen, then fonts, voices, WebGL vendor/renderer, and seeds are added separately. This creates fingerprints where the combination of properties never exists on any real device, which fingerprint consistency detectors (Pixelscan, Fingerprint.com) can identify as "masking."

**Specifically:** The WebGL vendor/renderer is sampled from `webgl.py` independently of the BrowserForge screen resolution and navigator platform. The font subset is generated from the OS font list independently of the GPU profile. A real device with a specific GPU (e.g., Intel Iris Xe) ships with specific screen resolutions and a specific font set — the synthetic generation doesn't enforce this correlation.

**Evidence from testing:** Running Camoufox v152.0.4-beta.28 (headed via Xvfb, no proxy, residential IP) against the Pixelscan fingerprint check:

- Browser: Firefox 152.0 on Ubuntu — ✅ (correctly spoofed)
- Location: Australia / Mango Hill — ✅ (GeoIP matched)
- Proxy: No proxy detected — ✅
- Bot check: No automated behavior detected — ✅
- **Fingerprint: Masking detected** — ❌ ("Your Browser Fingerprint is inconsistent")

The browser integrity, location, proxy, and bot detection checks all pass. Only fingerprint consistency fails — indicating the spoofed properties don't form a coherent device profile.

**Note:** This affects the BrowserForge synthetic path. The bundled `fingerprint-presets.json` and `fingerprint-presets-v150.json` presets (from real devices) should be internally consistent already.

### Screenshots:

Pixelscan results showing "Masking detected" with all other checks passing:

pixelscan-fingerprint-masking

### To Reproduce:

1. Launch Camoufox v152.0.4-beta.28 with default BrowserForge fingerprint generation (no preset)
2. Navigate to `https://pixelscan.net/fingerprint-check`
3. Observe "Masking detected" on the Fingerprint card while Browser, Location, Proxy, and Bot check all pass

```python

with Camoufox(headless="virtual") as browser:
    page = browser.new_page()
    page.goto("https://pixelscan.net/fingerprint-check")
    page.wait_for_timeout(30000)
    page.screenshot(path="pixelscan.png")
```

### Version:

```
Browser: v152.0.4-beta.28
BrowserForge: v1.2.4
OS: Linux x86_64 (headed via Xvfb :99, 1920x1080x24)
```

### Proposed solution

Add a consistency validation step in `generate_context_fingerprint()` (Python layer, `pythonlib/camoufox/fingerprints.py`) that ensures cross-property coherence when using BrowserForge synthetic generation:

1. **WebGL ↔ screen resolution:** When sampling WebGL vendor/renderer from `webgl.py`, restrict to GPU profiles that are commonly paired with the BrowserForge screen resolution (e.g., a 1366x768 screen shouldn't claim a high-end discrete GPU)
2. **WebGL ↔ OS platform:** Already partially handled (OS map in the WebGL sampling), but could be tighter
3. **Fonts ↔ OS:** Already handled via `_generate_random_font_subset(os_name)`, but marker font injection should match the claimed platform
4. **Voices ↔ OS:** Already handled via `_generate_random_voice_subset(os_name)`

The main gap is #1 — the WebGL profile and screen resolution are independently sampled and can form combinations that don't exist on real hardware. A lookup table mapping GPU profiles to common screen resolutions would close this.

This is a Python-only change in `fingerprints.py` — no C++/browser patch needed.

### LINKED ISSUE #731: Python voices config does not match browser schema and exposes host voices
The Python wrapper emits `voices` as strings, but Camoufox browser `v152.0.4-beta.28` expects complete voice objects. The browser skips the string entries and can expose the Linux host's native voices.

Related fingerprint-coherence discussion: https://github.com/daijro/camoufox/issues/729

## Environment

- Camoufox browser `v152.0.4-beta.28`
- `cloverlabs-camoufox==0.6.0`
- BrowserForge `1.2.4`
- Headed browser under Xvfb
- Linux host

## Python behavior

The bundled `voices.json` entries use a `Name:locale:type` representation.

The Python wrapper:

1. Loads those entries.
2. Keeps only the name.
3. Generates `List[str]`.
4. Places that list in `config["voices"]`.

The fallback path can place the original string records in the same property. Therefore, one property can receive two string formats.

The browser's `properties.json` declares only:

```json
{ "property": "voices", "type": "array" }
```

The Python validator checks that the value is a list. It cannot validate the element schema.

## Browser behavior

The browser source for this build requires every `voices` element to contain:

```json
{
  "lang": "string",
  "name": "string",
  "voiceUri": "string",
  "isDefault": true,
  "isLocalService": true
}
```

The parser skips a bare string or an incomplete object.

The build also supports:

```text
voices:blockIfNotDefined
```

When that option is false or absent, native host voices can still register.

## Reproduction

A bundled macOS real-device preset was launched on a Linux host.

With the wrapper-generated string configuration:

- The configured entries were skipped.
- `speechSynthesis.getVoices()` exposed 14,805 host Linux voices.

With structured objects and `voices:blockIfNotDefined=true`:

- The browser exposed only the intended configured voices.
- The host voice leak was removed.

No voice names, IP addresses, proxy details, or private infrastructure values are included here.

A Python-only shape check also demonstrates the mismatch:

```python
config = from_preset(preset, "152")
assert isinstance(config["voices"], list)
assert all(isinstance(voice, str) for voice in config["voices"])
```

## Expected behavior

- The Python package and browser must use one documented voice element schema.
- The Python package must preserve the locale and voice type needed to build that schema.
- Invalid elements must fail validation before browser launch.
- The browser must define whether an empty configured list blocks or permits native voices.
- A valid configured list must not expose unconfigured host voices.
- Wrapper/browser compatibility data must prevent installation of a pair with incompatible configuration schemas.

## Suggested acceptance criteria

- `from_preset` emits the documented voice object format.
- Every object contains `lang`, `name`, `voiceUri`, `isDefault`, and `isLocalService`.
- Element-level validation rejects strings and incomplete objects.
- A Linux-hosted browser using a macOS preset exposes no Linux host voices.
- Tests cover a valid list, an invalid string, an incomplete object, and an empty list.
- The expected interaction with `voices:blockIfNotDefined` is documented and tested.
- The supported Python-package/browser matrix includes the configuration-schema version.

This issue is separate from #729 because it has a specific type mismatch, reproduction, and acceptance boundary. The two issues are related because the leaked host voices make the resulting device profile internally inconsistent.

### LINKED ISSUE #741: Front TH not support
### Describe the bug:

Based on testing and real-world usage in .beta29, I encountered an issue with font rendering. When Thai characters were involved, the output was garbled—displaying unrecognizable, nonsensical characters. While English text rendered correctly, other languages ​​resulted in completely scrambled output.

but in .beta.28 is not issue

### Screenshots:

Image

### To Reproduce:

i guess UTF-8

### Version:

Python Packages                                                                 
  Camoufox                    v0.5.5                                            
  Browserforge                v1.2.4                                            
  Apify Fingerprints          v0.15.0                                           
  Playwright                  v1.60.0                                           
Browser                                                                         
  Active                      official/stable/152.0.4-beta.29                   
  Current browser             v152.0.4-beta.29                                  
  Build date                  Aug 20                                            
  SHA256                      620d33289b5d                                      
  Installed                   Yes                                               
  Latest in official/stable?  Yes                                               
  Last Sync                   2026-08-28 03:35                                  
GeoIP                                                                           
  Database                    MaxMind GeoLite2                                  
  Updated                     2026-08-28 03:35

JWriter20:
## Related Issue

Closes #741
Closes #731
Closes #729
Closes #653 (via the cherry-picked #742)

Also resolves the investigation in #733 — see "On #733" below; no code change was needed.

## Description

Five commits, each a single logical change. One carries upstream authorship: #742 (@LamerLink). #729 builds on ideas from #730 (@dyiapanis).

**`ad697a8` — #741: perturb glyph advances only, not positioning offsets**

`anti-font-fingerprinting.patch` fed a running total into each glyph's `x_offset`/`y_offset` as well as its advance:

```cpp
glyphPositions[i].x_advance += spacing;
glyphPositions[i].x_offset  += cumulativeOffset;   // <-- bug
cumulativeOffset += spacing;
```

`x_offset` is the GPOS positioning offset — the thing that parks a combining mark over its base. It contributes nothing to the measured width, so it adds nothing to the metric the randomisation exists to perturb, but layout never sees it either: `gfxTextRun` stores advances while painting honours the offsets. The running total therefore paints every glyph a further `spacing` past where layout placed it.

Latin hides this because Gecko's word cache shapes each space-delimited word separately, so the accumulator resets constantly. **Thai has no inter-word spaces**, so a whole paragraph is one shaping run and the error grows without bound — which is why the reporter saw English render fine and Thai come out mangled. Since the seed is randomised per launch, severity is a per-run lottery, which is why it looked like a beta.28 → beta.29 regression. It isn't: fonts, fontconfig and every font patch are byte-identical between those two releases.

**`160c806` — #731: make spoofed speech voices fail closed**

`nsSynthVoiceRegistry` only withheld the host's speech-dispatcher voices when the explicit `voices:blockIfNotDefined` flag was set. Nothing set it, so the host was suppressed only as a side effect of a non-empty spoofed list — and the Python layer built that list inside a bare `except Exception: pass`. Any path leaving the list empty or unset fell through to the host backend and exposed **14805 espeak-ng voices** under a fingerprint claiming macOS or Windows.

Closed at both layers: the browser now blocks whenever MaskConfig carries a `voices` array at all (including an empty one), `launch_options` pins the flag and degrades a generation failure to `[]`, and `validate_voices` rejects the bare-string / half-filled shapes `MVoices()` silently drops. With no `voices` key the browser still behaves like stock Firefox, so a bare binary is unaffected.

**`d6a806e` — #729: keep the WebGL renderer coherent with the screen**

The premise of a GPU↔screen lookup table has to survive one fact: **Firefox never reports the GPU it sees.** `dom/canvas/SanitizeRenderer.cpp` collapses every renderer string into one of ~11 representative device buckets before a page sees it — prefs `webgl.sanitize-unmasked-renderer` and `webgl.enable-renderer-query`, both default `true`; `resistFingerprinting` replaces the value with `"Mozilla"` outright. That file's own header comment gives the flavour:

> Narrow renderer string space down to representative replacements. E.g. `"GeForce RTX 3090"` => `"GeForce GTX 980"`

Two consequences drive the implementation.

*Raw model names can never match.* Every RTX, every Quadro M/P/V/T and every GeForce 900–7999 arrive as one string; `Intel(R) UHD Graphics 620` and `Mesa Intel(R) Iris(R) Xe Graphics` arrive as `Intel(R) HD Graphics` spellings. A table keyed on RTX / Quadro / RX / UHD / Iris / Mesa Intel matches nothing in `webgl_data.db` — all 33 rows are sanitized bucket strings. So renderers are reduced to their bucket first (`_renderer_bucket`), which also puts the ANGLE, nouveau and `/PCIe/SSE2` spellings of one GPU on one rule instead of three: the DB carries both `NVIDIA GeForce GTX 980, or similar` (proprietary driver) and `GeForce GTX 980, or similar` (nouveau, which loses the vendor prefix), and both must classify alike.

*A bucket carries almost no screen floor.* `GeForce GTX 980, or similar` is what a mobile GTX 1650 Max-Q reports, and `Radeon R9 200 Series, or similar` is what a Vega 8 budget laptop reports. 1366x768 laptops with discrete GPUs are ordinary hardware. The only claim true of *every* part behind a bucket is that discrete GPUs did not ship in netbooks, so that is the only claim encoded: the four discrete buckets require a screen larger than 1024x600. It is an **area**, not a width×height pair, because real panels do not dominate one another — 1280x800 and 1366x768 are both ordinary, and a per-axis floor taken from either rejects the other.

Three further properties:

- **Software rasterizers keep their natural rate.** A plain reject loop renormalises the surviving pool onto `llvmpipe` / WARP / SwiftShader; on a sub-floor screen that turns a 1.5% software rate into ~40% — trading a weak incoherence for the strongest VM/headless tell there is. So the first draw settles hardware-vs-software at the pool's own rate and is never resampled; rejection only ever runs within the hardware class.
- **Both entry points get the same treatment.** `generate_context_fingerprint()` — the per-context API #729's proposed solution names, and the one build-tester drives the browser through — sampled with a bare `sample_webgl()` and applied no floor. It now matches `launch_options()`.
- **Presets are never rewritten.** `raise_screen_to_modern_floor` lifts netbook-era BrowserForge geometry, but `_user_set_screen_window` is computed before a preset merges into the config, so the floor was rewriting real preset screens. Two bundled v150 presets genuinely report 736x414 and 960x540; both were being forced to 1366x768, breaking the very coherence #729 is about. The floor is now gated to synthetic draws.

From #730 (@dyiapanis): constrain the GPU to the screen rather than the screen to the GPU, so rejection sampling preserves `webgl_data.db`'s real OS weighting and leaves the already-reconciled geometry alone (#499); measure the screen-to-avail gap *before* mutating (#730 computed it after overwriting `screen.height`, turning a 1024x600 → 1080 bump into a 520px "taskbar"); and no Apple-M Retina floor, since Apple silicon also ships in the Mac mini and Mac Studio driving arbitrary external displays.

**`2b662a8` — Playwright 1.61+ support** (cherry-picked from #742 by @LamerLink, unmodified)

Included because the Playwright suite could not run at all without it — 1.62 produced 377 fixture errors, so this branch was untestable against current Playwright. Verified here (see below); #742 itself carried no test results.

**`17f873a` — bound the Playwright range instead of unpinning it**

#742 replaced `playwright = "<1.61"` with `"*"`. The breakage it fixes is real, but an unbounded range removes the tripwire rather than making the browser forward-compatible: `camoufox.server` imports `playwright._impl._driver.compute_driver_executable`, a private API with no compat guarantee, and `additions/juggler` is a fork of one Playwright vintage that every minor release is free to break again — 1.61 sending `viewport.isMobile` is exactly that, and it will recur. So the cap moves up to the tested-current 1.62.0 rather than away. No lower bound is added; this package has never carried one. `build-tester/requirements.txt` mirrors this pin by its own comment and moves with it.

Kept as a separate commit so the cherry-pick stays byte-identical to @LamerLink's — reading commit-by-commit, the pin therefore goes `<1.61` → `*` → `<1.63`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other

## Testing

Linux x86_64, built from this branch's own patch files (0 failed hunks). The built juggler in `dist/` is byte-identical to `additions/juggler/`, so the tested binary carries this branch.

**Before / after, measured on the built binary:**

| | before | after |
|---|---|---|
| #741 Thai painted-text overrun past its layout box (1x / 2x / 4x run length) | +4.90 / +9.80 / +20.60 px | **0.00 / 0.00 / 0.00 px** |
| #741 same measurement, Latin | 0.00 px | 0.00 px |
| #741 width perturbation (the actual fingerprinting signal) | 4 / 8 / 16 / 32 px | **unchanged** |
| #731 host voices exposed via `{"voices": []}` | 14805 | **0** |
| #731 host voices when generation raises | 14805 | **0** |
| #731 normal path | 115 | 115 |

**#729, 250 profiles per OS per entry point:**

| path | incoherent GPU/screen | min screen | software-renderer rate |
|---|---|---|---|
| `launch_options` win / mac / lin | 0 / 0 / 0 | 1600x900 / 1366x768 / 1440x900 | 1.2% / 0% / 12.4% |
| `generate_context_fingerprint` win / mac / lin | 0 / 0 / 0 | 1376x774 / 1366x768 / 1440x900 | 1.2% / 0% / 9.2% |

Software rates sit at the pool's natural baseline (1.5% win, 9.0% lin) within sampling noise — the constraint does not reweight the pool. Both sub-netbook v150 presets round-trip unchanged (736x414, 960x540).

Where the constraint actually binds is worth stating plainly: on the **default synthetic path it is a no-op**, because the netbook floor already guarantees ≥1366x768 and no discrete-GPU/netbook pair is reachable. It earns its place when the floor is skipped (user-pinned `screen.*`) or undone (headful on a genuinely small monitor):

| user-pinned 1024x600 | before | after |
|---|---|---|
| incoherent pairs, Windows | 52.2% | **0.0%** |
| incoherent pairs, Linux | 37.5% | **0.0%** |
| software-renderer rate, Windows | 2.00% | 2.25% |
| software-renderer rate, Linux | 9.55% | 10.10% |

**Playwright suite (`make tests`, under Xvfb)**

| binary | Playwright | result |
|---|---|---|
| stock beta.29 | 1.60 | 63 failed / 1033 passed |
| this branch (pre-#742) | 1.60 | 62 failed / 1034 passed |
| this branch (with #742) | 1.62.0 | **65 failed / 1031 passed / 107 skipped**, 0 fixture errors |
| stock or pre-#742 binary | 1.62.0 | **377 fixture errors** — suite cannot run |

NOTE: These failures are a result of the world isolation feature conflicting with the Playwright test suite, not real errors.

A display is required for a clean run: without one, the 7 `async/test_headful.py` tests fail purely for want of an X server (72 failed / 1024 passed), and they all pass under `xvfb-run`.

Set-difference against the stock baseline on 1.60: **0 failures unique to this branch.** All 62 also fail on stock beta.29; the baseline had one extra flaky one (`test_should_report_if_request_was_from_service_worker`).

The three extra failures on 1.62 were isolated to Playwright itself, not #742 — holding Playwright at 1.60 and varying only the binary, the +#742 build passes **6/6 websocket tests** and gives byte-identical tracing/network results to the pre-#742 build (same 2 failed / 60 passed, same test names). This matters because #742 makes `timestamp` a *required* protocol field on `webSocketFrameSent`/`webSocketFrameReceived`.

`pythonlib`: **156 passed**, including `tests/test_webgl_screen_consistency.py` (bucket reduction, the area floor, software-class preservation, both entry points, preset preservation) and fail-closed cases in `tests/test_voices.py`.

**Note on the uniqueness checks in the certificate below:** these flap run to run and fail on stock beta.29 too (observed 4/4, 3/4 and 2/4 on the same binary). Measured over 1500 trials, the chance that 3 generated profiles share a screen is **46.6% on Linux / 26.9% on macOS — identical with and without the change in `d6a806e`**. It is BrowserForge's 8-value screen pool, not a regression here. The screen floor is a no-op for the Linux and Windows pools (0/400 draws below it) and fires on 3.5% of macOS draws.

## Fingerprint Report

<details>
<summary>build-tester — Grade A, 1042/1048</summary>

```
╔════════════════════════════════════════════════════════════╗
║          CAMOUFOX BUILD VERIFICATION CERTIFICATE           ║
╠════════════════════════════════════════════════════════════╣
║  Grade: A     Score: 1042/1048     Profiles: 8             ║
║  Issued: 2026-08-29T19:39:19.248805Z                       ║
║  Status: FAILURES DETECTED                                 ║
╠════════════════════════════════════════════════════════════╣
║  SECTION RESULTS                                           ║
║  Automation Detection ..................... 32/32  [PASS]  ║
║  JS Engine ................................ 40/40  [PASS]  ║
║  Lie Detection .......................... 104/104  [PASS]  ║
║  Firefox APIs ............................. 88/88  [PASS]  ║
║  Cross-Signal ............................. 16/16  [PASS]  ║
║  CSS Fingerprint .......................... 64/64  [PASS]  ║
║  Math Engine .............................. 24/24  [PASS]  ║
║  Permissions .............................. 16/16  [PASS]  ║
║  Speech Voices ............................ 16/16  [PASS]  ║
║  Performance .............................. 16/16  [PASS]  ║
║  Intl Consistency ......................... 24/24  [PASS]  ║
║  Emoji ...................................... 8/8  [PASS]  ║
║  Canvas Noise ............................... 8/8  [PASS]  ║
║  WebGL Render ............................... 8/8  [PASS]  ║
║  Font Platform ............................ 16/16  [PASS]  ║
║  Audio Integrity .......................... 32/32  [PASS]  ║
║  Iframe Testing ........................... 24/24  [PASS]  ║
║  Headless Detection ....................... 80/80  [PASS]  ║
║  Trash Detection .......................... 40/40  [PASS]  ║
║  Font Environment ......................... 20/20  [PASS]  ║
║  Workers .................................. 64/64  [PASS]  ║
║  WebRTC ..................................... 8/8  [PASS]  ║
║  Stability .................................. 8/8  [PASS]  ║
║  Mac Uniqueness ............................. 4/4  [PASS]  ║
║  Linux Uniqueness ......................... 2/4  [2 FAIL]  ║
╠════════════════════════════════════════════════════════════╣
║  CROSS-PROFILE UNIQUENESS                                  ║
║  macOS  Audio:3/3  Canvas:3/3  TZ:3/3  Screen:3/3          ║
║  Linux  Audio:3/3  Canvas:2/3  TZ:3/3  Screen:2/3          ║
╠════════════════════════════════════════════════════════════╣
║  ID:   2c06acc0-8de1-4964-ade8-17c24d303491                ║
║  Hash: a295fcdbefe2ebfd99797129fe54b49b6587a0fed30b8453... ║
║  Sig:  2ace0e88b05a625468f0c19c1966fd3fb77d4debc4f4f754... ║
╚════════════════════════════════════════════════════════════╝
```
</details>

Run against this branch's `pythonlib` (`PYTHONPATH=../pythonlib`) so the certificate reflects the `generate_context_fingerprint` change, under `xvfb-run`.

Browser-level spot check across 6 launches (2 per emulated OS), reading `UNMASKED_RENDERER_WEBGL` and `speechSynthesis.getVoices()` from the page: every renderer is a real Gecko bucket coherent with its screen — `ANGLE (AMD, Radeon HD 3200 Graphics …)`, `ANGLE (NVIDIA, NVIDIA GeForce GTX 980 …)`, `Apple M1, or similar`, `Intel(R) HD Graphics, or similar`, and the bare-nouveau `NVIDIA GeForce GTX 980, or similar` — and voices come back 53 SAPI on Windows, 120–142 Apple on macOS, 131 on Linux, with no espeak leak.

service-tester not run — the checklist marks it as temporarily out of service.

## On #733

No code change. The reported leak does not reproduce on upstream: on beta.29 `page.evaluate()` sees **77** window properties (the isolated-world figure the reporter cites as the *good* case), page scripts cannot see its markers, and the `mw:` prefix correctly reaches the main world, so `main_world_eval` is not a no-op. None of `__pwInitScripts`, `__playwright_builtins__` or `__playwright__binding__` exist in the page's main world on any build tested — including with `cloverlabs-camoufox==0.6.0` itself.

The reporter's own data inverts their conclusion: their leaky run used **150.0.2-alpha.26** and the clean one beta.29. Confirmed against a v150 binary, where `evaluate()` sees **946** properties (main world) versus 76 on beta.29. It is a browser-version effect fixed by the isolated-world juggler work, not a package regression.

## Checklist

- [x] My changes are focused on a single logical change (one per commit)
- [ ] ~~Service tests pass~~ — marked out of service in the template
- [x] Build test passes — Grade A, 1042/1048, above the 1000 floor


KittisakPh:
@daijro pls check it and update pypi for us 💙🙏

KittisakPh:
@daijro My brotherrrrrrr pls check it

KittisakPh:
@daijro why you not review and merge this I don’t know what happened but you can update README.md but not prove merge on Camoufox issue

JWriter20:
> @daijro why you not review and merge this I don’t know what happened but you can update README.md but not prove merge on Camoufox issue

Relax @KittisakPh, daijro largely doesn't want to be involved with camoufox because he's got other work going on. He's not obligated to maintain a past project.

I'll take a final look over it and make sure everything's good, then make it a pre-release.

JWriter20:
@daijro Gave a final pass through and fixed a few things, looks good to merge now.

PR #746, opened 2026-08-31
release: v152.0.4-beta.30 + pythonlib 0.5.6, with a Playwright-keyed browser floor

JWriter20:
Releases the work merged since beta.29 (#743, and #745 now that it has landed).

Rewritten since the first review: the flat `MIN_VERSION` this PR originally used has been replaced by a floor keyed on the resolved Playwright version, and auditing that turned up three further latent bugs, all fixed here.

## Why a browser floor is needed at all

`pyproject.toml` on main raises the Playwright cap to `<1.63`, which is only safe against a browser whose `Protocol.js` accepts the viewport fields 1.61 began sending. Measured against real builds:

| | playwright 1.60 | 1.61 | 1.62 |
|---|---|---|---|
| **beta.29** | OK | FAIL | FAIL |
| **beta.30** | OK | — | OK |

The failure is `Protocol error (Browser.setDefaultViewport)` on *every* `new_context()`. Old pythonlib against a new browser is safe — the new schema fields are `t.Optional`.

## Why the floor is conditional, not flat

The incompatibility is two-dimensional — it needs Playwright >= 1.61 **and** a browser < beta.30 — but `MIN_VERSION` only knows about the browser. A flat floor has to assume the worst Playwright, which means:

- every 0.5.6 user re-downloads the browser, including the majority on <1.61 who are in no danger;
- installs **pinned** to an older build lose the pin, and **prerelease/alpha** users are moved off their channel (every `alpha.*` sorts below `beta.30`);
- the library cannot run *at all* until the matching browser release is published, making the PyPI-after-release ordering load-bearing.

Keyed on Playwright instead, only the pairing that actually breaks moves:

```
playwright <1.61   -> floor alpha.1  -> every install kept
playwright >=1.61  -> floor beta.30  -> below-beta.30 installs upgraded
version unreadable -> floor alpha.1  -> kept (a spurious forced re-download is
                                        worse than leaving a working install)
```

## Three latent bugs the floor exposed

All the same shape: code downstream of the version check that had never executed, because an `alpha.1` floor never rejected anything.

1. **`camoufox_path()` raised `FileNotFoundError` instead of fetching.** It probed `INSTALL_DIR/version.json`, which only the pre-multiversion flat layout ever wrote. Any raised floor would have crashed every existing user rather than upgrading them.
2. **Unbounded recursion.** The function ended in `return camoufox_path()`. When the newest *published* build is still below the floor, `install()` is a no-op and that tail recursed — measured **990 fetch attempts before `RecursionError`**, exhausting the unauthenticated GitHub rate limit (60/hr) on the way. Now re-checks once and raises `UnsupportedVersion` naming the requirement.
3. **`executable_path` was ignored for bundle resolution.** `get_env_vars()`/`_generate_fontconfig()` read the bundled fontconfig and fonts through `get_path()` — the managed install — even when the caller supplied their own binary (`_load_properties()` already honoured it for `properties.json`). Previously this silently mixed one build's fonts into another's launch; with a live floor it became fatal, failing every launch while the caller held a perfectly good binary. Found by stress-testing, and it would have hit anyone running their own build or a CI-downloaded artifact.

## Verification — clean build, full runthrough

Built from scratch on this branch (`make dir` + Linux x86_64), then:

| Check | Result |
|---|---|
| Stress suite, 18 launch configurations | **18/18** |
| `tests/patches/*` | **7/7** |
| pythonlib | **172 passed** |
| vermin (3.8 gate) | pass |
| build-tester | **405/408 — identical to shipped beta.29** |
| Vendored Playwright suite (1.62, with display) | **6 failed / 1088 passed** |

The stress suite asserts observable behaviour through the DOM (page-world state is invisible to `page.evaluate` by design): OS spoofing reaching the page, explicit config passthrough, a user-set 1024x600 screen surviving the #729 floor, voices failing closed, all four init-script quadrants, `navigator.webdriver`, and invalid config being rejected loudly rather than degrading.

The 6 remaining suite failures are all pre-existing and characterised: 3 are stale vendored assertions against newer Playwright (the `Page.waitForEventInfo` -> `__waitInfo__` rename, verified in `_waiter.py`, and a websocket error-string change); 2 fail **0/5 on stock Playwright Firefox** too, so they are test defects rather than browser bugs; and 1 (`test_frame_goto_should_continue_after_client_redirect`) is a genuine camoufox `networkidle` issue that predates this work (vanilla 17/17, camoufox 9/17) and deserves its own issue. build-tester's 3 are the pre-existing `selfDestruct.setTimezone`, byte-identical to beta.29.

## Backwards compatibility for 0.5.6

Not all from this PR, but shipping in this release:

- **`voices` is now strictly validated** (#743). `voices=["Alex:en-US:local"]` — the string form in `camoufox/voices.json` — now raises `InvalidPropertyType`. Previously accepted and silently ignored, which was the #731 bug; converting a silent failure to a loud one is the point, but it *is* breaking for anyone passing that shape.
- **Generated fingerprints change** (#729): sub-1366x768 screens are lifted and WebGL is sampled to match the screen. Explicit `screen.*` values and presets are respected (covered by a stress case); defaults are not.
- **Forced browser upgrades are now limited** to users on Playwright >= 1.61, rather than everyone.

## Release order

Still browser-first — merge, tag `v152.0.4-beta.30`, let `build.yml` publish the assets, verify, then dispatch `publish-pypi.yml`. The conditional floor means an ordering slip is no longer catastrophic (users on <1.61 are unaffected either way), but it remains the right order.

Note this repo has no PR CI (`build.yml` and `publish-pypi.yml` are both tag/dispatch only), so the table above is the whole verification story. Linux x86_64 only — windows and macos are first exercised by the tag build.

### DIFF
--- a/pythonlib/camoufox/__version__.py
+++ b/pythonlib/camoufox/__version__.py
@@ -8,8 +8,26 @@ class CONSTRAINTS:
     The minimum and maximum supported versions of the Camoufox browser.
     """
 
-    MIN_VERSION = 'beta.30'
+    MIN_VERSION = 'alpha.1'
     MAX_VERSION = '1'
+
+    # The browser floor is conditional on the resolved Playwright, not fixed.
+    #
+    # Each entry is (playwright_version, required_browser_build): from that
+    # Playwright on, the browser must be at least that build. 1.61 began
+    # sending viewport isMobile/screenSize in Browser.setDefaultViewport and
+    # Page.setViewportSize; beta.30 is the first build whose Protocol.js schema
+    # accepts them. Below that pairing every new_context() dies with
+    # "Protocol error (Browser.setDefaultViewport)". Measured: 1.60 works on
+    # beta.29 and beta.30; 1.61 and 1.62 fail on beta.29 and pass on beta.30.
+    #
+    # A flat MIN_VERSION cannot express this. It only knows about the browser,
+    # so to stay safe it has to assume the worst Playwright and force *every*
+    # user to re-download -- including the majority on <1.61, who are in no
+    # danger -- and it leaves the library unusable until the matching browser
+    # release is published. Keyed on Playwright, only the users who would
+    # actually break get moved.
+    PLAYWRIGHT_BROWSER_FLOORS = (((1, 61), 'beta.30'),)
 
     @staticmethod
     def as_range() -> str:
--- a/pythonlib/camoufox/__version__.py
+++ b/pythonlib/camoufox/__version__.py
@@ -8,7 +8,7 @@ class CONSTRAINTS:
     The minimum and maximum supported versions of the Camoufox browser.
     """
 
-    MIN_VERSION = 'alpha.1'
+    MIN_VERSION = 'beta.30'
     MAX_VERSION = '1'
 
     @staticmethod
--- a/pythonlib/camoufox/pkgman.py
+++ b/pythonlib/camoufox/pkgman.py
@@ -755,6 +755,22 @@ def installed_verstr() -> str:
     return Version.from_path(active).full_string
 
 
+def _root_install_supported() -> bool:
+    """
+    Whether INSTALL_DIR's root holds a supported build.
+
+    Only the pre-multiversion flat layout wrote version.json at the root; the
+    versioned layout keeps it under browsers/<repo>/<version>/. A missing root
+    version.json means "no legacy install here", so the caller should fall
+    through to a fetch rather than raise. The alpha.1 floor masked this: no
+    install was ever unsupported, so this branch was never reached.
+    """
+    try:
+        return Version.from_path().is_supported()
+    except FileNotFoundError:
+        return False
+
+
 def camoufox_path(download_if_missing: bool = True) -> Path:
     """
     Full path to the active camoufox folder
@@ -787,7 +803,7 @@ def camoufox_path(download_if_missing: bool = True) -> Path:
                 f"{active_display} is not installed. " f"Please run `camoufox fetch` to install."
             )
 
-    elif os.path.exists(INSTALL_DIR) and Version.from_path().is_supported():
+    elif os.path.exists(INSTALL_DIR) and _root_install_supported():
         return INSTALL_DIR
 
     else:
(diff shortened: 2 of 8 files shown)


JWriter20:
## Added: a second latent bug the floor exposes

Found while auditing backwards compatibility. `camoufox_path()` ended in `return camoufox_path()` after a fetch. When the newest **published** build is still below `MIN_VERSION`, `install()` is a no-op (`"already installed"`) and that tail recurses — measured **990 fetch attempts before `RecursionError`**, having exhausted the unauthenticated GitHub API rate limit (60/hr) on the way. I hit the real 403 by accident before reproducing it deterministically offline.

This is reachable in two ways:

1. **The PyPI-before-browser window.** If 0.5.6 reaches PyPI before the beta.30 assets are live, this is what every user gets — not the clean `MissingRelease` I described above.
2. **Permanently**, for anyone using a `repos.yml` browser source that does not carry beta.30 (`coryking/camoufox`, `JWriter20/camoufox`).

Same class as the `FileNotFoundError` in the first commit: the `alpha.1` floor meant this branch was never reached, so it rotted unnoticed. Fixed by re-checking after the fetch and raising `UnsupportedVersion` naming the required minimum. Regression test added; 163 pythonlib tests pass.

## Backwards-compatibility notes for 0.5.6

Not all of these come from this PR, but they ship in this release and are worth stating:

- **`voices` config is now strictly validated** (from #743). `voices=["Alex:en-US:local"]` — the string form that ships in `camoufox/voices.json` — now raises `InvalidPropertyType`. Previously accepted and silently ignored, which was the #731 bug; this converts a silent failure into a loud one, but it *is* a breaking change for anyone passing that shape.
- **The floor forces an upgrade.** Every install below beta.30 is replaced: pinned installs lose their pin, and prerelease/alpha users are moved to beta.30 (all `alpha.*` sort below it). Offline or air-gapped installs stop working, since the upgrade cannot be fetched.
- **Generated fingerprints change** (#729): screens below 1366×768 are lifted, and WebGL is sampled to match the screen. Explicit `screen.*` settings and presets are respected; defaults are not.

A conditional floor keyed on the resolved Playwright version would avoid forcing the upgrade on users who are on Playwright <1.61 and in no danger, and would shrink the ordering hazard to nothing. Still happy to do that if wanted.

PR #750, opened 2026-09-01
fix(fingerprint): seal the window.setXxx() setters before page script (#749)

### LINKED ISSUE #749: 13 fingerprint setters are exposed to page script on window (callable, enumerable)
The `set*` fingerprint setters are declared in `dom/webidl/Window.webidl`, so they are members of the **Window interface** — present on the page's own `window`, enumerable, and callable by ordinary page script.

Measured on the shipped `152.0.4-beta.30` Linux x86_64 build.

## What a page sees

An inline `<script>` on a served page — i.e. what a *site* sees, not what automation sees:

```
visible to page: 13/14

['setFontSpacingSeed', 'setAudioFingerprintSeed', 'setTimezone', 'setScreenDimensions',
 'setScreenColorDepth', 'setNavigatorPlatform', 'setNavigatorOscpu',
 'setNavigatorHardwareConcurrency', 'setWebGLVendor', 'setWebGLRenderer', 'setFontList',
 'setSpeechVoices', 'setWebRTCIPv4']
```

Identical whether launched through the camoufox Python package or through raw Playwright.

They are not merely reachable by name:

```
typeof                   function
'setTimezone' in window  true
own_descriptor           {writable: true, enumerable: true, configurable: true}
getOwnPropertyNames      includes it
for...in over window     finds it
toString                 function setTimezone() { [native code] }
```

Because they are **enumerable own properties**, a generic "diff this window against a reference Firefox" script finds all 13 without knowing any of their names.

## They are also callable, and they work

```js
navigator.hardwareConcurrency            // 8
window.setNavigatorHardwareConcurrency(999)   // no throw
navigator.hardwareConcurrency            // 999
```

So a page can both detect camoufox with certainty and rewrite the values it is shown.

## Why the existing protection does not cover this

Two mechanisms are easy to conflate, and only one applies:

* **World isolation (#707)** hides JS that camoufox adds *inside its sandbox*. These setters are not injected JS — they are on the Window interface, which belongs to the page. Isolation cannot hide them, and never could.
* **The `Func=` guard plus `DisableFunction()` self-destruct** is the mechanism that is supposed to hide them. It only fires **when a setter is called**.

That is the gap. The normal Python flow applies the fingerprint through `CAMOU_CONFIG` in C++ and does not call these setters at all — `fingerprints.py` only emits a `setTimezone` call, and only when a timezone is explicit. Nothing calls them, so nothing self-destructs, so they stay on `window` for the page's whole lifetime.

Verified that the C++ path is sufficient on its own — raw Playwright, no Python layer, so no init script exists:

```
CAMOU_CONFIG timezone only  ->  page tz = Asia/Tokyo
```

## Two smaller defects found alongside

**`TimezoneManager::DisableFunction()` is never called.** It is defined correctly and does the `PutBool`, but nothing invokes it, unlike its siblings:

```
speech-voices-spoofing.patch   defines :65   called :171   OK
font-list-spoofing.patch       defines :73   called :182   OK
timezone-spoofing.patch        defines :54   never called  BUG
```

So `setTimezone` stays exposed even in the one flow that *does* call it — the `JS_DeleteProperty` in `SetTimezone` is futile, because the `Func=` guard still returns true and the WebIDL resolve hook re-adds the property.

**`setCanvasSeed` does not exist.** It is in build-tester's `SELF_DESTRUCT_FUNCTIONS` list but appears nowhere in `patches/` or `additions/`, so that check passes vacuously.

## Why build-tester reports this as healthy

build-tester scores 405/408 with only `selfDestruct.setTimezone` failing, which reads as "14 of 15 fine". That is an artifact of its own flow: it *calls every setter itself* to build per-context profiles, which triggers each `DisableFunction` before it checks. It is measuring a configuration real users do not run.

## Reproduction

Serve a page containing:

```html
<!doctype html><body><script>
  const S = ['setFontSpacingSeed','setAudioFingerprintSeed','setTimezone','setScreenDimensions',
             'setScreenColorDepth','setNavigatorPlatform','setNavigatorOscpu',
             'setNavigatorHardwareConcurrency','setWebGLVendor','setWebGLRenderer','setFontList',
             'setSpeechVoices','setWebRTCIPv4'];
  document.title = S.filter(n => typeof window[n] !== 'undefined').length + '/13 exposed';
</script></body>
```

then `page.goto(url)` and read `page.title()`. Reading via `page.evaluate()` will **not** show this correctly — that runs in the isolated world, which is the same blind spot as #738.

## Directions

Rather than making the self-destruct fire more reliably, it seems better not to expose these to content at all:

* **Chrome-gate them** — require a chrome/addon caller in the `Func=` guard, the way `shadow-root-bypass.patch` uses `Func="Document::IsCallerChromeOrAddon"`. Content never sees them regardless of whether they have been called. Needs checking that camoufox's own callers still qualify — the isolated-world sandbox holds an expanded principal, not a system principal.
* **Drop the content-facing API** where `CAMOU_CONFIG` already covers it, and use juggler's chrome-side `Browser.setTimezoneOverride` (`BrowserHandler.js:257`) for the per-context case.

The `DisableFunction` call is worth adding regardless, but on its own it fixes one symptom of the 13.

Happy to send a PR once there is a view on which direction is preferred — the chrome-gating change touches every manager, so it is worth agreeing the approach first.

JWriter20:
Fixes #749.

## The problem

Camoufox applies a per-context fingerprint through fifteen helpers on `window` — `setNavigatorPlatform()`, `setTimezone()`, `setWebGLRenderer()` and the rest. Each removed itself from `window` when called, so the set was only ever cleaned up as far as a given fingerprint happened to reach:

- a value the config leaves alone never calls its setter. `setTimezone` (no timezone configured) and `setWebRTCIPv6` (never emitted by any code path) stayed on `window` for **every** context;
- a launch that registers no init script at all left **all fifteen**. That is `Camoufox()` followed by `browser.new_page()`, the documented default, where the fingerprint arrives through `CAMOU_CONFIG` in C++ and nothing ever touches a setter;
- so did launching the binary as a plain browser, with no juggler running.

Page script could not only see them but call them:

```js
navigator.hardwareConcurrency            // 8
window.setNavigatorHardwareConcurrency(999)
navigator.hardwareConcurrency            // 999
```

Fifteen `window` properties that no other Firefox build has is a sharper fingerprint than any of the values they were hiding.

## The fix

A window is now created **sealed**, and juggler opens it for exactly as long as init scripts are running:

```
window created -> sealed -> unseal -> init scripts apply the fingerprint
                         -> seal   -> page script runs, setters gone
```

Sealed is the default so the failure is closed: a path that does not go through juggler never offers the setters, instead of offering them permanently to whatever loads.

**Per window, not per browsing context.** A context-wide flag — which is effectively what the existing self-destruct was — fires on the initial `about:blank` and leaves every later navigation unable to apply anything, so both contexts come up with the launch-level fingerprint instead of their own. That case is a permanent assertion in the test.

**Deleting the names is not enough on its own.** They are handed out lazily by the DOM resolve hook, so a deleted name comes straight back (`delete window.alert` behaves the same way). The `Func=` guards consult the seal, which is what makes a deletion stick; the delete clears what the generated init script's `typeof w.setX === "function"` probe has already materialised.

## Why the suite did not catch it

`build-tester` scored this **A** throughout. Its checklist omits `setNavigatorUserAgent` and `setWebRTCIPv6` — the two that leaked on every context — while asserting the absence of `setCanvasSeed`, which has never existed anywhere in the tree and so passed vacuously. Corrected here.

## Measured

Linux x86_64, this build vs the shipped `152.0.4-beta.29`. Counts are camoufox setters visible to **page** script (an inline `<script>` in the served document — `page.evaluate()` runs in the isolated world and cannot see this):

| path | before | after |
|---|---|---|
| `Camoufox()` + `browser.new_context()` | 15 | **0** |
| context with an init script | 2 | **0** |
| `page.add_init_script()` after `new_page()` | 15 | **0** |
| binary launched directly, no juggler | 15 | **0** |

Fingerprints still apply exactly as requested in every case (`Win32/5120`, `MacIntel/1512`, `MacIntel/2816` all matched their generated config).

## Tests

- `tests/patches/fingerprint-setter-seal.py` (new) — five scenarios, all pass; covers both the leak and the per-window regression.
- `tests/patches` — 7/7 pass.
- `build-tester` — 962/962, Grade A. Its one Linux-uniqueness miss is a preset-pool collision unrelated to this change (30 draws yield only 7 distinct Linux screen sizes, `1600x900` nine times); worth its own issue.

https://claude.ai/code/session_018i4WFSwSGLNH2d7wVVKFjQ

### DIFF
--- a/additions/juggler/content/FrameTree.js
+++ b/additions/juggler/content/FrameTree.js
@@ -692,6 +692,12 @@ class Frame {
     // global means state written by evaluate() on one page is still there on
     // the next -- every other world starts empty per document.
     this._masterSandbox = null;
+    // Camoufox: open the window.setXxx() fingerprint setters for the init
+    // scripts below. A window is created sealed, so this is the only moment
+    // they exist -- see nsGlobalWindowInner::CamouSettersSealed.
+    const camouInnerWindowId = this.domWindow().windowGlobalChild.innerWindowId;
+    ChromeUtils.camouUnsealFingerprintSetters(camouInnerWindowId);
+
     this._createIsolatedContext('', true);
     for (const [name, world] of this._frameTree._isolatedWorlds) {
       if (name)
@@ -703,6 +709,19 @@ class Frame {
       for (const script of world._scriptsToEvaluateOnNewDocument)
         this._evaluateInitScript(executionContext, script);
     }
+
+    // Close them again. The init scripts have had their turn and page script
+    // has not run yet, so this is the last moment at which nobody untrusted has
+    // been able to look.
+    //
+    // Trusting each setter to remove itself when called only ever covered the
+    // setters a given fingerprint happened to set. A value the config left
+    // alone (no timezone, no IPv6) left its setter sitting on window, and a
+    // launch that registers no init script at all -- Camoufox() +
+    // browser.new_page(), the documented default -- left all fifteen. Fifteen
+    // window properties no other Firefox has is a sharper fingerprint than
+    // anything they were hiding.
+    ChromeUtils.camouSealFingerprintSetters(camouInnerWindowId);
 
     const url = this.domWindow().location?.href;
     if (url === 'about:blank' && !this._url) {
--- a/build-tester/src/lib/checks/index.ts
+++ b/build-tester/src/lib/checks/index.ts
@@ -9,18 +9,19 @@ export interface PhaseResult {
 const SELF_DESTRUCT_FUNCTIONS = [
   "setFontSpacingSeed",
   "setAudioFingerprintSeed",
-  "setCanvasSeed",
   "setTimezone",
   "setScreenDimensions",
   "setScreenColorDepth",
   "setNavigatorPlatform",
   "setNavigatorOscpu",
   "setNavigatorHardwareConcurrency",
+  "setNavigatorUserAgent",
   "setWebGLVendor",
   "setWebGLRenderer",
   "setFontList",
   "setSpeechVoices",
   "setWebRTCIPv4",
+  "setWebRTCIPv6",
 ];
 
 function runSelfDestructChecks(): Record<string, CheckResult> {
--- a/patches/window-setter-seal.patch
+++ b/patches/window-setter-seal.patch
@@ -1,0 +1,359 @@
+diff --git a/dom/base/AudioFingerprintManager.cpp b/dom/base/AudioFingerprintManager.cpp
+index 6ec4dc5265..d8a5a5aac0 100644
+--- a/dom/base/AudioFingerprintManager.cpp
++++ b/dom/base/AudioFingerprintManager.cpp
+@@ -91,6 +91,9 @@ AudioFingerprintManager::IsFunctionEnabledForWebIDL(JSContext* aCx, JSObject* aO
+   if (!win) {
+     return false;
+   }
++  // Camoufox: sealed once this window's init scripts have run, so page
++  // script never sees the setter (FrameTree -> camouSealFingerprintSetters).
++  if (win->CamouSettersSealed()) return false;
+ 
+   uint32_t userContextId = 0;
+   if (BrowsingContext* bc = win->GetBrowsingContext()) {
+diff --git a/dom/base/ChromeUtils.cpp b/dom/base/ChromeUtils.cpp
+index 8657d3282b..d8889edb6a 100644
+--- a/dom/base/ChromeUtils.cpp
++++ b/dom/base/ChromeUtils.cpp
+@@ -6,6 +6,10 @@
+ #include "MaskConfig.hpp"
+ #include "MouseTrajectories.hpp"
+ 
++// Camoufox: for CamouSealFingerprintSetters below.
++#include "mozilla/dom/ScriptSettings.h"
++#include "nsGlobalWindowInner.h"
++
+ #include "JSOracleParent.h"
+ #include "ThirdPartyUtil.h"
+ #include "VsyncSource.h"
+@@ -2781,6 +2785,74 @@ void ChromeUtils::CamouGetMouseTrajectory(GlobalObject& aGlobal, long aFromX,
+   aPoints.AppendElements(flattenedPoints.data(), flattenedPoints.size());
+ }
+ 
++// Camoufox: every per-context fingerprint setter, in the order the managers
++// above are listed. Kept next to the seal because the seal is the only place
++// that needs the names as a set -- each manager otherwise knows only its own.
++static const char* const kCamouFingerprintSetters[] = {
++    "setNavigatorPlatform", "setNavigatorOscpu",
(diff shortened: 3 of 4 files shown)


KittisakPh:
@JWriter20 Why not merged

JWriter20:
> @JWriter20 Why not merged

Made that commit right before I left on a 22 mile hiking/camping trip. Driving back home now, still need to look it over again before merging

KittisakPh:
@JWriter20 Ok bro thx you very much and can u release at browser and pypi cuz now stuck pre-release 3 days ago

JWriter20:
> @JWriter20 Ok bro thx you very much and can u release at browser and pypi cuz now stuck pre-release 3 days ago

You should still be able to use the pre-release no (both the browser and pypi)? I typically like to keep it at a pre-release for a few days so people can notify about issues before it becomes the main release.

KittisakPh:
@JWriter20 ok sir i waill wating for release this 3 days already thx you for your patian hope you not fotget it

KittisakPh:
@JWriter20 i have some question for use camoufox 
handless True or Flase is better for using cuz in https://github.com/techinz/browsers-benchmark

it show if use handless False is bad more handless True 
not show browser is better for show browser right or is the same 
can you guide me for the best way to use 

thx you very much pls guide me

JWriter20:
> @JWriter20 i have some question for use camoufox handless True or Flase is better for using cuz in https://github.com/techinz/browsers-benchmark
> 
> it show if use handless False is bad more handless True not show browser is better for show browser right or is the same can you guide me for the best way to use
> 
> thx you very much pls guide me

Screenshot From 2026-09-03 20-33-39

Screenshot From 2026-09-03 20-58-52


I just ran the benchmark myself for all 3 OSes, each with fingerprints matching the actual operating system, using the Pypi [0.5.6b1](https://pypi.org/project/camoufox/0.5.6b1/) and [v152.0.4-beta.30](https://github.com/daijro/camoufox/releases/tag/v152.0.4-beta.30)   All cases scored 90% or 100%, the recorded entry in his benchmark is just wrong, or outdated (maybe the old 135 version). 

For full stealth, you should always run headed or with xvfb, and match the fingerprint of your os to your machine, I may add full stealth for headless at some point, but I have heard from others in the community that covering all of the cases is such a pain in the ass that its not worth it. However if it is more convenient for you, test out headless with your target and if it works then great, no need to use headed.

KittisakPh:
@JWriter20 Thx you very much bro i'm applicate it
but
pls update Pypi [0.5.6b1](https://pypi.org/project/camoufox/0.5.6b1/) and [v152.0.4-beta.30](https://github.com/daijro/camoufox/releases/tag/v152.0.4-beta.30) for release

 i'm very excited to useeeeeee it pls transfer pre-release to release plsss.

Issue #744, opened 2026-08-29
UserAgent and Browser version mismatch detected


mubashirhussainrao:
### Website detecting Camoufox:
https://www.browserscan.net/browser-checker
### Screenshots:

<img width="1273" height="515" alt="Image" src="https://github.com/user-attachments/assets/ed301c11-6a34-45e9-8ece-4073b62d7b08" />

### To Reproduce:

        self.browser_manager = AsyncCamoufox(
            proxy=prox,
            humanize=0.8,
            geoip=True,
            fingerprint_preset=True,
            enable_cache=True,
            block_images=True,
            disable_coop=True,
            os='linux',
        )
        self.browser = await self.browser_manager.start()
        self.page = await self.browser.new_page()
        await self.page.goto('https://www.browserscan.net/browser-checker')

### Other questions:

These questions will help me diagnose the issue:

1. Are you using a proxy?
YES

3. Is Camoufox detected randomly or every time?
everytime

4. What OS are you using?
am using macOs and fingerprint of linux but changing fingerprint OS have no improvement

### Version:

Python Packages                                                                   
  Camoufox                    v0.5.5                                              
  Browserforge                v1.2.4                                              
  Apify Fingerprints          v0.15.0                                             
  Playwright                  v1.55.0                                             
Browser                                                                           
  Active                      official/stable/152.0.4-beta.29                     
  Current browser             v152.0.4-beta.29                                    
  Build date                  Aug 20                                              
  SHA256                      eeada250746e                                        
  Installed                   Yes                                                 
  Latest in official/stable?  Yes                                                 
  Last Sync                   2026-08-27 07:07      

also tried browser:  official/stable/152.0.4-beta.28

PR #754, opened 2026-09-04
fix(pythonlib): derive navigator.appVersion from the preset's platform

### LINKED ISSUE #753: fingerprint_preset leaks the host OS through navigator.appVersion
### Describe the bug:

`from_preset()` sets `navigator.userAgent`, `navigator.platform` and `navigator.oscpu` from the captured device, but never `navigator.appVersion`. Firefox reports `appVersion` as `"5.0 (<OS token>)"`, so an unset value falls through to the **host's** own — and any page reading two properties sees them contradict each other.

Measured on `152.0.4-beta.29`, macOS host, headless, `geoip=False`:

| launch | `navigator.platform` | `navigator.appVersion` |
| --- | --- | --- |
| `os="linux"` (generated) | `Linux x86_64` | `5.0 (X11)` |
| `os="windows"` (generated) | `Win32` | `5.0 (Windows)` |
| `os="macos"` (generated) | `MacIntel` | `5.0 (Macintosh)` |
| `os="linux", fingerprint_preset=True` | `Linux x86_64` | **`5.0 (Macintosh)`** ← the host |

The generated (browserforge) path is coherent on all three, which is why this only shows on the preset path.

The same launch, read together as a page would:

```
ua         : Mozilla/5.0 (X11; Linux x86_64; rv:152.0) Gecko/20100101 Firefox/152.0
oscpu      : Linux x86_64
platform   : Linux x86_64
appVersion : 5.0 (Macintosh)
```

`navigator.appVersion` is already declared in `settings/properties.json`, so nothing in the browser needs to change — only the Python layer never populates it.

### To Reproduce:

```python
import asyncio
from camoufox import AsyncCamoufox

async def main():
    async with AsyncCamoufox(headless=True, os="linux", geoip=False,
                             fingerprint_preset=True) as browser:
        await page.set_content("<body></body>")
        print(await page.evaluate(
            "() => [navigator.platform, navigator.appVersion]"
        ))

asyncio.run(main())
```

On a macOS host this prints `['Linux x86_64', '5.0 (Macintosh)']`. On a Linux host, try `os="windows"`.

### Other questions:

1. Are you using a proxy? No — `geoip=False`, nothing network-dependent.
2. Is it random or every time? Every time, on every preset, because the value is never set.
3. What OS are you using? macOS host; the leak is the host's OS whatever that is.

### Version:

```
Browser   official/stable/152.0.4-beta.29
Camoufox  0.5.4 (also present at HEAD — `appVersion` appears nowhere in pythonlib/camoufox/fingerprints.py)
```

### Note

This may well be what #744 is seeing — that report is `fingerprint_preset=True` with `os='linux'` on a macOS host, which is exactly this configuration, and browserscan's "browser version mismatch" check reads `appVersion`. I could not confirm from the screenshot there, so I am filing this separately rather than assuming.

Happy to send a PR — I have a one-function fix and a regression test that fails without it.

polyackiy:
## Related Issue

Closes #753 (and possibly #744 — see the note at the end).

## Description

`from_preset()` sets `navigator.userAgent`, `navigator.platform` and `navigator.oscpu` from the captured device, but never `navigator.appVersion`. Firefox reports `appVersion` as `"5.0 (<OS token>)"`, so an unset value falls through to the **host's** own — and a page reading two properties sees them contradict each other.

Measured on `152.0.4-beta.29`, macOS host, headless, `geoip=False`:

| launch | `navigator.platform` | `navigator.appVersion` |
| --- | --- | --- |
| `os="linux"` (generated) | `Linux x86_64` | `5.0 (X11)` |
| `os="windows"` (generated) | `Win32` | `5.0 (Windows)` |
| `os="macos"` (generated) | `MacIntel` | `5.0 (Macintosh)` |
| `os="linux", fingerprint_preset=True` | `Linux x86_64` | **`5.0 (Macintosh)`** ← the host |

The generated path is already coherent, which is why this only shows on the preset path.

The value is derived exactly the way `oscpu` already is a few lines above — from the platform the preset carries. Two deliberate choices:

- **A preset that ships its own `appVersion` keeps it.** The captured device's value is the real one; the derivation is only a fallback.
- **A platform Firefox does not run on is left alone** rather than guessed at, so an unexpected preset cannot be given a token that no real Firefox emits.

`navigator.appVersion` is already declared in `settings/properties.json`, so nothing in the browser changes.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other

## Testing

`pythonlib/tests/test_preset_appversion.py` — six cases, in the style of `test_webgl_screen_consistency.py`: the token each platform implies, that `appVersion` and `userAgent` name the same system, that a preset's own value wins, and that an unknown platform is untouched.

- **Without the patch: 5 of the 6 fail.** (The sixth — unknown platform — passes either way, by design.)
- **With it: 6 pass.**
- **The rest of the Python suite is unaffected: `180 passed, 4 skipped`.**

End to end against a real browser, `152.0.4-beta.29`, running this branch's `pythonlib` rather than a monkeypatch — the same probe as the table above:

```
os=linux, from preset   platform=Linux x86_64   appVersion=5.0 (X11)     (was 5.0 (Macintosh))
```

## Fingerprint Report

Being upfront rather than pasting something I did not run:

- **service-tester** — not run. It requires real proxies, which I do not have. This change is confined to `from_preset()` and is covered by the unit tests above plus the live browser check; if you would like the service-tester output before merging, I am happy to wait for someone who can run it.
- **build-tester** — not run either, and I do not think it applies: it bypasses the Python package by design, and this patch is entirely inside `pythonlib`. The binary is unchanged.

Say the word if you would rather have both regardless and I will find a way.

---

**On #744:** that report is `fingerprint_preset=True` with `os='linux'` on a macOS host — exactly this configuration — and browserscan's "browser version mismatch" check reads `appVersion`. I could not confirm it from the screenshot, so I filed #753 separately instead of assuming, and this PR closes that one.

### DIFF
--- a/pythonlib/camoufox/fingerprints.py
+++ b/pythonlib/camoufox/fingerprints.py
@@ -738,6 +738,33 @@ def get_random_preset(
     return choice(candidates)  # nosec
 
 
+# Tokens that name the machine rather than the platform: Firefox leaves every one
+# of them out of appVersion.
+_APP_VERSION_DROPPED = ('Win64', 'x64', 'Mobile', 'Tablet')
+
+
+def _app_version_from_user_agent(user_agent: str) -> Optional[str]:
+    """The appVersion Firefox reports for a browser sending this user agent.
+
+    "5.0 (<OS tokens>)": the parenthesised part of the UA without the
+    architecture, the Gecko revision, or the Windows build number.
+    """
+    block = re.match(r'Mozilla/5\.0 \(([^)]*)\)', user_agent or '')
+    if not block:
+        return None
+    kept = []
+    for token in (part.strip() for part in block.group(1).split(';')):
+        if (
+            token.startswith('rv:')
+            or token in _APP_VERSION_DROPPED
+            or token.startswith('Linux ')
+            or token.startswith('Intel Mac OS X')
+        ):
+            continue
+        kept.append('Windows' if token.startswith('Windows') else token)
+    return f"5.0 ({'; '.join(kept)})" if kept else None
+
+
 def from_preset(preset: Dict, ff_version: Optional[str] = None) -> Dict[str, Any]:
     """
     Convert a real fingerprint preset to CAMOU_CONFIG format.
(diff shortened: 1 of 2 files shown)


JWriter20:
Thanks @polyackiy. Landed via #756 (merged as 35b45c1), commit fff2c73, authorship intact. Fixes #753.

Issue #225, opened 2025-03-04
Mouse movement bug


gaotian19981104:
"After I started the browser and added the mouse movement trajectory, I found a bug that appeared after version 134. When the program executes the trajectory, if my real mouse enters the browser window, the mouse trajectory will automatically stop and the code will stop running."

monolithed:
@gaotian19981104  did you manage to find a workaround?

rubenvereecken:
Hey — we've been hitting the same thing. Tracked it down to `activateAndRun` in `TargetRegistry.js`: it serialises all mouse dispatches on a module-global promise chain (`globalTabAndWindowActivationChain`). If the C++ `juggler-mouse-event-hit-renderer` observer notification doesn't fire for one event, the await never resolves and every subsequent mouse op queues behind it forever — process-wide, across all contexts.

Keyboard works fine because `Page.dispatchKeyEvent` bypasses the chain.

We've got a workaround in our fork: wrap the inner block in a short `Promise.race` timeout. On timeout, the rejection flows through the existing `.catch(() => {})`, chain advances, next mouse dispatch runs clean. Turns permanent mouse death into a single retryable error.

One gotcha: chrome-privileged JS doesn't have `setTimeout`/`clearTimeout` in global scope — need to import from `Timer.sys.mjs`.

Interesting: in our testing the timeout never actually fires — the extra event-loop tick from `Promise.race` + `setTimeout` seems to shift timing enough to avoid the race entirely. But without it, we reliably deadlock within a few hundred trajectories on a plain `example.com` page (macOS arm64, headed, 146.0.1).

Haven't opened a PR for this yet because I can't pin down the actual root cause — only contain it. Likely affects #279 and #553 too — same chain, same deadlock, different entry point.

JWriter20:
Fixed by #680, now on `main`.

@rubenvereecken's diagnosis in this thread was exactly right and is what made this findable — `activateAndRun` in `TargetRegistry.js` serializes every mouse dispatch on a process-global promise chain, and a single missing `juggler-mouse-event-hit-renderer` notification wedges all subsequent input, process-wide. Thank you for that write-up.

## What was actually producing the missing ack

A mousemove at exactly `x==width` or `y==height` fires as an exit event rather than `eMouseMove`, so no ack is ever generated. That was fixed for the **requested endpoint** back in 9270618 (`>=` rather than `>`).

[...]

PR #755, opened 2026-09-04
fix(juggler): make the input-dispatch deadlock structurally unreachable (#751, #752)

### LINKED ISSUE #751: humanize: trajectory points at x==0 / y==0 still deadlock the global input chain (#225 fix covers only the far edges)
### Summary

The exact-edge bounds guard added in #680 (beta.28) fixes only the **far** edges
(`>= width`, `>= height`). The **near** edges are still let through, because the
check is `< 0` rather than `<= 0`:

```js
// PageHandler.js, humanize trajectory
if (currentX < 0 || currentY < 0 || currentX >= boundingBox.width || currentY >= boundingBox.height)
//   ^^^^^^^^^^^^^^^^^^^^^^^^^^ x==0 / y==0 pass                    ^^^^^^^^^^ fixed in #680
  continue;
```

A trajectory point at exactly `x == 0` or `y == 0` is dispatched, fires as an exit
event rather than `eMouseMove`, produces no `hit-renderer` ack, and wedges the
process-global activation chain — the same deadlock #225/#680 describe, just from
the opposite side of the viewport.

### Reproduction / evidence

Instrumented `PageHandler.js` (logging around `camouGetMouseTrajectory` and each
`sendOne`) on `v152.0.4-beta.28`, real site, cold start:

```
[HUMZ] traj-start from=(0,0) to=(667,302) box=1294x676.6
[HUMZ] traj-got n=114
[HUMZ] send i=2 (31,0)      <- y == 0; dispatched, no ack, never returns
```

No `ack` line follows. The click times out; every later input in that process hangs.

This also explains two symptoms that were confusing before:

- **The page observes zero `mousemove` events during the hang** — the point fired as
  an exit event, so a `mousemove` listener never sees it.
- **It clusters on the session's first pointer input.** `_lastTrackedPos` starts at
  `{x: 0, y: 0}` (`PageHandler.js:86`), so the first trajectory always departs from
  the top-left corner; with the +/-80px knot boundary from `MouseTrajectories.hpp`
  that reliably produces points on `x == 0` / `y == 0`. Once a move lands on an
  interior point, later trajectories never touch the near edges.

Cold-start hang rate on our workload: **1 in 3** before the change below,
**0 in 20** after.

### Suggested fix

Make the guard symmetric:

```js
if (currentX <= 0 || currentY <= 0 || currentX >= boundingBox.width || currentY >= boundingBox.height)
  continue;
```

`tests/patches/humanize-edge-deadlock.py` passes on beta.28 as-is because its
`EDGE_TARGETS` only probe the right/bottom edges. Adding a near-edge target (e.g.
`(1, 350)` / `(500, 1)`, or simply asserting the very first move of a session)
would have caught this.

Optionally, initialising `_lastTrackedPos` to an interior point (as
`coryking/camoufox` does with a random in-viewport position) would additionally
remove the systematic corner departure — though on that fork the guard is still
the older strict-`>` form, so both edges remain reachable there.

### Environment

- browser: `v152.0.4-beta.28` (verified identical humanize block in `v152.0.4-beta.30`)
- pythonlib: `0.5.4`
- headless, `humanize=True`

### LINKED ISSUE #752: Mouse click is not working
### Describe the bug:

I did not know why I was not able to pass cloudflare turnstile, but apparently mouse click does nothing, it does not fire the click event, tried it in different methods to click, same issue. Mouse move works, only click is the issue. I tried it with humanaze and no humanize.

### Version:
Also tried beta.28 and beta.29

Python Packages                                                        
  Camoufox                        v0.5.5                               
  Browserforge                    v1.2.4                               
  Apify Fingerprints              v0.15.0                              
  Playwright                      v1.60.0                              
Browser                                                                
  Active                          official/prerelease                  
  Current browser                 v152.0.4-beta.30

JWriter20:
## Related Issue

Closes #752
Closes #751

## Description

`Mouse.move()` / `click()` at certain viewport coordinates permanently wedges **all** input dispatch in the browser process — every tab, no error, no CPU use, no diagnostic. This is the fifth report of the same family (#225 and its relatives were the earlier ones), so this PR fixes the instance *and* removes the class.

### The bug

`sendOne` dispatched at `eventY + boundingBox.top`. A relative `y == 0` lands at exactly `boundingBox.top`, which is the content area's first half-covered device row. The widget rounds to a whole row, and wherever `Math.round(top) < top` that row still belongs to chrome — so the event hit-tests outside the content widget and fires as an *exit* event instead of `eMouseMove`. No `juggler-mouse-event-hit-renderer` ack is produced.

`EventWatcher.ensureEvent()` then waits **forever**, holding `globalTabAndWindowActivationChain` — a module-global promise that serializes every input dispatch in the process. One bad coordinate deadlocks the whole browser.

It is deterministic per spoofed OS, because the chrome height differs:

| spoofed OS | `boundingBox.top` | rounds to | result |
|---|---|---|---|
| windows | 51.4 | 51 | hangs |
| macos | 53.1 | 53 | hangs |
| linux | 56.5 | 57 | fine |

That per-OS split is why this kept getting reported and kept looking unreproducible — `humanize-mouse-trajectory.py` happened to pin `os="linux"`, the one immune fingerprint. `x == 0` was never affected because `boundingBox.left` is a whole `0`.

### The fix — three layers

Guarding each call site is what we've done four times already, and it keeps breaking because the guard has to be remembered at every new dispatch site. Instead:

**1. One chokepoint.** New `additions/juggler/input/MouseDispatch.js` owns relative→absolute conversion, the bounds predicate, and the ack wait. Absolute coordinates are clamped into the last whole device pixel inside the element on all four sides. `PageHandler.js` **loses 92 lines and gains 22** — its three ad-hoc bounds checks and every raw dispatch call are gone, so the vendored file's diff against upstream juggler gets *smaller*. Wheel events now go through the same conversion, so a wheel at `y == 0` no longer scrolls the tab strip.

**2. Bounded waits.** `EventWatcher.ensureEventWithin()` replaces the unbounded wait: 5s, then drop the event and log type/coordinate/rect. A trajectory abandons its remaining points after the first undelivered one. `activateAndRun()` gets a 30s backstop so the chain can never be held forever by anything, acks included.

The 5s figure is measured, not guessed — over 1000+ dispatches the ack is p50 0ms / p99 1ms / max 12ms idle, but **max 2849ms behind a 3s synchronous script**, because the ack is delivered from the content main thread and inherits any block on it. A 1s deadline would silently drop legitimate input on slow pages. A wall-clock budget across a whole trajectory was rejected for the same reason: it would false-fire on exactly the slow pages the deadline exists to tolerate.

**3. Enforcement.** `scripts/check-input-dispatch.py` fails if `jugglerSendMouseEvent`, `sendWheelEvent`, `jugglerEventId` or `boundingBox.left/top` appear anywhere outside the chokepoint, wired into a new `.github/workflows/lint.yml`. Two documented content-process exemptions: `PageAgent.js` (drag, already content-relative) and `FrameTree.js` (the ack *producer*, which the lint itself flagged). `docs/input-dispatch.md` states the invariant: *a synthesized input event whose ack we await must reach the content renderer — and when it does not, we must stop waiting.*

Note `build.yml` only runs on tags, so nothing was checking PRs before this.

### A fifth instance, found by the new sweep

On its first run, on a build that already had the near-edge fix, `mouse-boundary-sweep.py` found another one. `boundingBox.height` is consistently **0.5 CSS px less** than the `innerHeight` the page reports, so the page's last row is half-covered. With the box at `1920x977.5 +0+56.5`, relative `y == 977` — `innerHeight - 1`, well inside the viewport as far as the page is concerned, and somewhere Playwright will happily click — dispatches at 1033.5, rounds to 1034, and the content ends at 1034. Deterministic 4/4, pre-existing, and it deadlocks a stock build. Fixed by the same symmetric clamp.

It's the first one in this family found before a user hit it. The ack backstop is what makes the sweep possible at all: on stock, the first bad coordinate kills the run.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other

## Testing

All results below are from a **cold build** — `rm -rf` of the source dir, re-extract, patches re-applied, `mach build` + `mach package` — not an incremental tree.

**`tests/patches/` — 14/14 pass, zero non-zero exits.** Three are new:

- `near-edge-mouse-deadlock.py` — `{windows, macos, linux}` × `{(31,0), (0,0), (500,0)}`, plus a humanized endpoint and cold-page first-click runs. Fails on stock at the first move.
- `mouse-boundary-sweep.py` — `{0, 1, w//2, w-2, w-1} × {0, 1, h//2, h-2, h-1}` across 3 spoofed OSes × humanize on/off (150 coordinates, 6 scenarios), asserting each point is both acked and observed by the page.
- `input-ack-backstop.py` — blocks the content main thread for 40s and asserts the move returns bounded. **Measured 5.0s.** It uses a genuine late ack rather than a test-only hook in shipped code.

To reproduce (three prerequisites that are easy to miss — `tests/patches/` is *not* run by `make tests`, which is `pytest async/` only):

```bash
make stage-fonts
# needs a venv with the camoufox package; tests/setup-venv.sh does not build one.
# build-tester/.venv works:
CAMOUFOX_EXECUTABLE_PATH=<path>/dist/bin/camoufox-bin \
  build-tester/.venv/bin/python tests/patches/mouse-boundary-sweep.py
```

**`async/` (the `make tests` suite) — 14 failures, all pre-existing.** Verified by A/B rather than assumed: swapping the juggler files to `upstream/main`'s on the same binary — which is the entirety of this change, since it's JS only — fails 13 of the same 14. Nothing passes on stock and fails with this branch. The one delta goes the other way (`test_page_clock.py::TestWhileRunning::test_should_pause` fails on stock, passes here — a timing test, so likely flake rather than a fix). Seven of the fourteen are `test_headful.py` under the runner's default `--headless`.

**Packaging.** `jar.mn` adds the first *new file* juggler has gained, so it was verified end-to-end rather than by inspection: `chrome/juggler/content/input/MouseDispatch.js` (10052 bytes) is present in the packaged `omni.ja`. `package-manifest.in:207` packages juggler as a tree, not a file list, so no manifest change is needed.

**Lint.** `input-dispatch: ok -- 16 files scanned`. Verified in both directions — passes clean, and fails when a dispatch site is injected.

**Not covered:** the 30s `activateAndRun` backstop has no automated test. Exercising it needs a stall that isn't an ack, which would mean a hook in shipped code; it rests on prototype measurement only.

## Fingerprint Report

Service tester is marked out of service in the template, so build tester only. Certificate text rather than a screenshot:

<details>
<summary>Fingerprint report — build-tester, Grade A, 962/962</summary>

```
╔════════════════════════════════════════════════════════════╗
║          CAMOUFOX BUILD VERIFICATION CERTIFICATE           ║
╠════════════════════════════════════════════════════════════╣
║  Grade: A     Score: 962/962     Profiles: 8               ║
║  Issued: 2026-09-04T18:23:36.074883Z                       ║
║  Status: ALL PASS                                          ║
╠════════════════════════════════════════════════════════════╣
║  SECTION RESULTS                                           ║
║  Automation Detection ..................... 32/32  [PASS]  ║
║  JS Engine ................................ 40/40  [PASS]  ║
║  Lie Detection .......................... 104/104  [PASS]  ║
║  Firefox APIs ............................. 88/88  [PASS]  ║
║  Cross-Signal ............................. 16/16  [PASS]  ║
║  CSS Fingerprint .......................... 64/64  [PASS]  ║
║  Math Engine .............................. 24/24  [PASS]  ║
║  Permissions .............................. 16/16  [PASS]  ║
║  Speech Voices ............................ 16/16  [PASS]  ║
║  Performance .............................. 16/16  [PASS]  ║
║  Intl Consistency ......................... 24/24  [PASS]  ║
║  Emoji ...................................... 8/8  [PASS]  ║
║  Canvas Noise ............................... 8/8  [PASS]  ║
║  WebGL Render ............................... 8/8  [PASS]  ║
║  Font Platform ............................ 16/16  [PASS]  ║
║  Audio Integrity .......................... 32/32  [PASS]  ║
║  Iframe Testing ........................... 24/24  [PASS]  ║
║  Headless Detection ....................... 80/80  [PASS]  ║
║  Trash Detection .......................... 40/40  [PASS]  ║
║  Font Environment ......................... 20/20  [PASS]  ║
║  Workers .................................. 64/64  [PASS]  ║
║  WebRTC ..................................... 8/8  [PASS]  ║
║  Stability .................................. 8/8  [PASS]  ║
║  Mac Uniqueness ............................. 4/4  [PASS]  ║
║  Linux Uniqueness ......................... 3/4  [1 FAIL]  ║
╠════════════════════════════════════════════════════════════╣
║  CROSS-PROFILE UNIQUENESS                                  ║
║  macOS  Audio:3/3  Canvas:3/3  TZ:3/3  Screen:3/3          ║
║  Linux  Audio:3/3  Canvas:2/3  TZ:3/3  Screen:3/3          ║
╠════════════════════════════════════════════════════════════╣
║  ID:   6bb02b09-2dec-4f17-96d2-c930732958a2                ║
╚════════════════════════════════════════════════════════════╝
```

</details>

The `Linux Uniqueness 3/4` row is worth a look but is **not** from this PR — cross-profile uniqueness isn't part of the 962 score, and across three runs the collision moves between signals and platforms (run 1: macOS Canvas + Linux Screen; run 2: clean; run 3: Linux Canvas). Two of three runs hit a *canvas* collision, though, which looks like a real pre-existing weakness in per-profile canvas seeding. Happy to open a separate issue.

## Checklist

- [ ] ~~Service tests pass~~ — marked out of service in the template
- [x] Build test passes — Grade A, 962/962, certificate above (text rather than screenshot)

## Two things noticed in passing, not fixed here

1. **`make build` on Linux silently cross-compiles for macOS.** `scripts/patch.py:239` defaults to `macos, arm64` when `BUILD_TARGET` is unset, so a bare `make build` on a Linux host produces `obj-aarch64-apple-darwin` and **exits 0** while `xcrun` and `rust-objcopy` fail in warnings. It looks like a successful build until you notice the obj directory name. Defaulting to the host triple, or refusing to guess, would save someone an hour.
2. **`tests/patches/` has no runner.** It isn't part of `make tests`, the files are hyphenated so pytest won't collect them, and they need both `make stage-fonts` and a venv containing the `camoufox` package that `tests/setup-venv.sh` doesn't create. Wiring it into a make target would make it much harder to skip.

## Open design questions

- Should an undelivered event surface to the Playwright client rather than only the browser log? With warn-and-continue, a user not running `DEBUG=pw:browser` sees a click that silently does nothing — which is exactly what #752 looked like.
- Should `globalTabAndWindowActivationChain` be scoped per window instead of per process? Process-global is why one bad coordinate in one tab freezes every tab.

https://claude.ai/code/session_01GQgHHGRXNp29jr4xQjK7iv

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,21 @@
+name: Lint
+
+# Static checks only -- no browser build, so these can gate every pull request.
+# The build workflow runs on tags and takes ~40 minutes; nothing was checking
+# pull requests before this.
+on:
+  pull_request:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+jobs:
+  input-dispatch:
+    name: Synthesized input goes through one chokepoint
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - run: python3 scripts/check-input-dispatch.py
--- a/additions/juggler/Helper.js
+++ b/additions/juggler/Helper.js
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const uuidGen = Cc["@mozilla.org/uuid-generator;1"].getService(Ci.nsIUUIDGenerator);
+const {setTimeout, clearTimeout} = ChromeUtils.importESModule("resource://gre/modules/Timer.sys.mjs");
 
 export class Helper {
   decorateAsEventEmitter(objectToDecorate) {
@@ -172,6 +173,8 @@ export class Helper {
 
 const helper = new Helper();
 
+const kEventTimedOut = Symbol('event-timed-out');
+
 export class EventWatcher {
   constructor(receiver, eventNames, pendingEventWatchers = new Set()) {
     this._pendingEventWatchers = pendingEventWatchers;
@@ -199,6 +202,35 @@ export class EventWatcher {
       if (result)
         return result;
       await new Promise((resolve, reject) => this._pendingPromises.push({resolve, reject}));
+    }
+  }
+
+  /**
+   * Like ensureEvent, but gives up after timeoutMs and resolves null instead of
+   * waiting forever.
+   *
+   * Callers awaiting an ack for synthesized input must use this. Input dispatch
+   * is serialized on activateAndRun()'s process-global promise chain, so an ack
+   * that never arrives does not merely lose one event -- it wedges every later
+   * input event in the process, in every tab, permanently, at 0% CPU with no
+   * diagnostic. Four shipped deadlocks (#225, #677, #751, #752) were all that
+   * failure. Bounding the wait is what makes the fifth one a log line.
+   */
+  async ensureEventWithin(aEventName, timeoutMs, predicate) {
+    const pending = this.ensureEvent(aEventName, predicate);
+    // Whichever promise loses the race stays pending until dispose() rejects
+    // it; swallow that so a timed-out wait never surfaces as an unhandled
+    // rejection in chrome JS.
+    pending.catch(() => {});
+    let timer;
+    const timedOut = new Promise(resolve => {
+      timer = setTimeout(() => resolve(kEventTimedOut), timeoutMs);
+    });
+    try {
+      const result = await Promise.race([pending, timedOut]);
+      return result === kEventTimedOut ? null : result;
+    } finally {
+      clearTimeout(timer);
     }
   }
 
--- a/additions/juggler/TargetRegistry.js
+++ b/additions/juggler/TargetRegistry.js
@@ -11,6 +11,23 @@ const {AppConstants} = ChromeUtils.importESModule("resource://gre/modules/AppCon
 // scripts), so the screencast tick has to import them explicitly.
 const {setTimeout, clearTimeout} = ChromeUtils.importESModule("resource://gre/modules/Timer.sys.mjs");
 
+// Last-resort bound on how long one callback may occupy the process-global
+// activation chain below. The chain must always advance: a callback that never
+// returns wedges every later input event in every tab, permanently.
+//
+// The per-ack deadline in MouseDispatch.js covers the await that has actually
+// caused all four shipped deadlocks, but it is one of several unbounded waits
+// reachable from a single slot -- apz-repaints-flushed, TabSwitchDone below,
+// the drag path's juggler-drag-finalized and dragover waits, and the
+// cross-process dispatchDragEvent sends all have the same shape. None of them
+// has failed yet. Bounding only the wait that has already bitten us is the
+// posture that produced those four fixes, so bound the slot itself too.
+//
+// Sized as a backstop, not a tuning knob: with a 5s ack deadline a legitimate
+// worst-case input slot approaches 10s, so this must sit well clear of that.
+const kActivationSlotBudgetMs = 30000;
+const kSlotExpired = Symbol('activation-slot-expired');
+
 const Cr = Components.results;
 
 const helper = new Helper();
@@ -512,7 +529,18 @@ export class PageTarget {
       const notificationsPopup = muteNotificationsPopup ? this._linkedBrowser?.ownerDocument.getElementById('notification-popup') : null;
       notificationsPopup?.style.setProperty('pointer-events', 'none');
       try {
-        await callback();
+        let timer;
+        const expired = new Promise(resolve => {
+          timer = setTimeout(() => resolve(kSlotExpired), kActivationSlotBudgetMs);
+        });
+        try {
+          if (await Promise.race([callback(), expired]) === kSlotExpired) {
+            dump(`[juggler] WARN activation-chain slot exceeded ` +
+                 `${kActivationSlotBudgetMs}ms; advancing the chain without it\n`);
+          }
+        } finally {
+          clearTimeout(timer);
+        }
       } finally {
         notificationsPopup?.style.removeProperty('pointer-events');
       }
--- a/additions/juggler/input/MouseDispatch.js
+++ b/additions/juggler/input/MouseDispatch.js
@@ -1,0 +1,229 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use strict";
+
+/**
+ * The one place in the parent process that dispatches synthesized mouse input.
+ *
+ * It exists because the alternative did not work. Between 2026-04 and 2026-09,
+ * four separate deadlocks shipped -- exact-edge coordinates (#225), humanized
+ * trajectory points that bypassed the endpoint's guard (#677/#225), a
+ * zero-displacement move, and the top-edge row (#751/#752) -- each fixed by
+ * adding one more coordinate guard at one more call site. The guards were
+ * correct; the arithmetic behind them was copied per call site, so every new
+ * dispatch site was a fresh chance to get it wrong, and every mistake cost the
+ * whole browser process.
+ *
+ * THE INVARIANT
+ *   A synthesized input event whose ack we await must reach the content
+ *   renderer -- and when it does not, we must stop waiting.
+ *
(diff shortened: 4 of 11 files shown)


KittisakPh:
@JWriter20 Why not merge bro

JWriter20:
Thanks @JWriter20. Landed via #756 (merged as 35b45c1). The chokepoint refactor and the chrome/content boundary fix are both in, and `tests/patches/` carries four guards for it — each verified to reproduce the deadlock on a stock build before passing here.

Issue #737, opened 2026-08-24
Search service fails to initialize: bundled `search-config-overrides-v2` dump is missing `recordType`


hekman316:
## Summary

`SearchService.#init()` throws on every launch, on a clean profile, in both
`152.0.4-beta.28` and `152.0.4-beta.29`. The browser ends up with **no search
engines at all**, and — because the urlbar's heuristic result depends on the
search service — the address bar drop-down shows **nothing**: no history, no
autofill, no bookmarks. Typing a word navigates to it as a hostname instead.

## Error

```
console.debug: SearchService: "init"
console.error: SearchService: "#init: failure initializing search:"
    (new Error("reason: JSON error: missing field `recordType` at line 1 column 114",
     "moz-src:///toolkit/components/uniffi-bindgen-gecko-js/components/generated/RustSearch.sys.mjs", 2681))
```

## Cause

The bundled Remote Settings dump

```
omni.ja → defaults/settings/main/search-config-overrides-v2.json
```

holds one record whose keys are:

```
clickUrl, id, identifier, last_modified, partnerCode, schema, telemetrySuffix, urls
```

There is no `recordType`, which the Rust `SearchEngineSelector` requires. For
comparison, every one of the 157 records in the sibling `search-config-v2.json`
does carry `recordType`, so only the overrides dump is stale.

## Steps to reproduce

```bash
camoufox fetch official/152.0.4-beta.29
# fresh profile, search logging on
printf 'user_pref("browser.search.log", true);\n' > /tmp/prof/user.js
camoufox -profile /tmp/prof -headless -no-remote about:blank
```

The error appears within a second of startup. `search.json.mozlz4` is never
written, and the profile has no engines.

## Impact

- No search from the address bar; a typed word is treated as a hostname.
- The urlbar drop-down is empty — history and autofill are unreachable even
  though `places.sqlite` is populated and `browser.urlbar.suggest.history` is
  `true`. All urlbar providers register fine (`UrlbarProviderPlaces`,
  `UrlbarProviderAutofill`, `UrlbarProviderHeuristicFallback`); the results just
  never get built.
- Adding an engine through `policies.json` (`SearchEngines.Add`) does not help,
  and neither does a WebExtension declaring `chrome_settings_overrides.
  search_provider` — both register through the same dead service.

## Suggested fix

Refresh the bundled `search-config-overrides-v2.json` from Remote Settings so
its record carries `recordType`, or ship `{"data": []}` — the overrides
[...]

PR #756, opened 2026-09-05
Integration branch: eleven open PRs built and verified together on FF152

JWriter20:
Integration branch: open PRs built and verified together on FF152. History is linear — one
commit per contribution, original authorship intact, no merge commits — so merging this lands
each contributor's work as their own commit. The individual PRs it supersedes are listed below
and should be closed against it.

Verified merge-ready: `mergeable_state: clean`, 0 commits behind `main`, 22 ahead, no conflicts.
Still marked draft; flip to "Ready for review" to merge.

## Included (8 PRs)

| PR | Author | Area |
| --- | --- | --- |
| #755 | @JWriter20 | input-dispatch deadlock (#751, #752) |
| #754 | @polyackiy | `navigator.appVersion` leaks host OS on the preset path (#753) |
| #703 | @pratyush618 | keep browser chrome out of the spoofed font list (#695) |
| #697 | @nxbdev | `navigator.maxTouchPoints` from config (#696) — **completed, see below** |
| #716 | @luandhgt | verify sha256 of release assets; TLS on IP lookup (#715) |
| #722 | @codechrl | re-download addons with a missing manifest (#308) |
| #562 | @rubenvereecken | `media:spoof_codecs` (#558) — needed a schema fix |
| #624 | @koopatroopa787 | Windows `run_tests.ps1` only — **mozglue commit dropped** |
| #315 | @datawookie | headless screen size (#242) — needed a follow-up fix |

## Dropped after review

- **#704** — a link to the submitter's own project, placed inside the "Camoufox's approach"
  section. Closed.
- **#735 (`navigator.buildID`)** — premise is inverted. `20181001000000` is Firefox's
  `LEGACY_BUILD_ID` (`nsRFPService.h:62`), and stock `Navigator::GetBuildID` returns it for
  every page not on `https://*.mozilla.org`. No Camoufox patch touches that function. Deriving
  a period-correct buildID would make Camoufox the only Firefox reporting a non-legacy value —
  a novel tell rather than a fixed one.
- **#624's mozglue commit** — @GleckusZeroFive reports the external-manifest approach does not
  resolve the SxS failure on a clean Windows 11 box, because Windows *merges* an external
  manifest with the embedded one rather than replacing it (also #650). Never updated since.
  The `run_tests.ps1` half is kept.

## Fixes added on top

- **`media-codec-spoofing.patch` did not apply to FF152.** Its `#include "MaskConfig.hpp"` hunk
  targeted an include block FF152 reordered; since patches apply in sequence, it aborted the
  whole run. Retargeted.
- **`media:spoof_codecs` was never declared**, so `validate_config()` dropped it and the
  documented usage did nothing. Declared, plus a guard scanning for `MaskConfig::Get*("key")`
  reads missing from the schema.
- **#315's guard flip needed a call-site fix.** `get_screen_cons(headless or has_display(env))`
  meant a headful run on a real display read as headless and lost its monitor bound — undoing
  2266f27/#499.
- **`no-search-engines.patch` broke the search service outright (#737).** Its stub returned a
  search-config **v1** record, but the selector is the Rust `SearchEngineSelector`, which
  deserializes v2 — where `recordType` is the serde enum tag, so a record without it aborts the
  whole document. `SearchService.#init()` threw ``missing field `recordType` `` on every launch
  in beta.28 through beta.30, leaving no search service at all, and with it no urlbar heuristic
  result. The issue blames a stale bundled `search-config-overrides-v2` dump; the reporter
  retracted that after measuring, and both bundled dumps are fine as shipped — the overrides
  dump is never even parsed, because `setSearchConfig` throws first. The failing document is the
  stub itself: serialising it gives exactly 116 characters with `}` at column 114, which is
  where serde reports the missing field. It now returns v2 records — one **inert** engine whose
  search URL is loopback, plus `defaultEngines` and `engineOrders`. An empty configuration is
  not an option: `getEngineConfiguration()` rejects `[]`, and `SearchSettings` refuses to write
  without an engine. Verified: 0 `recordType` errors, `Completed #init`, `search.json.mozlz4`
  written, and the only engine present is the inert `None` that `policies.json` already defines.

- **Dropped three stale `.bak` files.** `additions/juggler/TargetRegistry.js.bak` was being
  copied into every source tree by `copy-additions.sh` (`cp -r ../additions/* .`) — it never
  reached `omni.ja`, so it did not ship to users, but it had no reason to be in the build.
  The two `.patch.bak` files were never picked up by `list_patches()`, which globs `*.patch`,
  so both had been inert since they landed. Patch discovery still finds 56 patches.

## #697 completed

Measured against a real Windows touchscreen (Dell XPS 15 9510, Firefox 152.0). All 16 signals
now match — the three that did not are fixed, and the other 13 are unchanged:

| signal | real touchscreen | before | now |
| --- | --- | --- | --- |
| `(any-pointer: coarse)` | true | false | **true** |
| `window.TouchEvent` | true | false | **true** |
| `window.Touch` | true | false | **true** |

- `force-default-pointer.patch` regains its `aID` branch, so `coarse` joins the **any-pointer**
  set, and only when `maxTouchPoints > 0`. The primary pointer stays `Fine | Hover`: a
  touchscreen laptop still drives its trackpad, and `(pointer: coarse)` would claim a phone
  while the desktop UA said otherwise. The host `LookAndFeel` value is still not consulted, so
  the capability set cannot vary with the build machine.
- The interfaces come from moving `TouchEvent::PrefEnabled` only, never `LegacyAPIEnabled`.
  `dom.w3c_touch_events.legacy_apis.enabled` is false everywhere but Android — which is exactly
  why a real Windows touchscreen laptop exposes `TouchEvent` and `Touch` while
  `'ontouchstart' in window` is false. Switching touch on wholesale would be *more* detectable
  than doing nothing. `maxTouchPoints=0` stays byte-identical to before.
- `mobile-fingerprint-spoofing.patch` → `touchscreen-fingerprint-spoofing.patch`: the rationale
  is the ordinary Windows touchscreen laptop, not a phone. The rename reorders it after
  `navigator-spoofing.patch`, so its `Navigator.cpp` hunk now applies at an offset — verified
  clean, no rejects.
- A launch-time warning fires whenever the option is set, separately from the blanket
  `navigator.` warning, because the effects reach past navigator into the CSS pointer media
  queries and the touch interfaces.
- `tests/patches/touchscreen-digitizer.py` checks all 16 recorded signals plus a
  `maxTouchPoints=0` control. It fails on a binary built without the change (12/16) and passes
  with it.

**This changes the fingerprint of ordinary preset launches, not just manual config.** 22 of the
bundled presets ship `maxTouchPoints` above zero (18 at 10, one at 5, three at 1) — all of them
`Win32`, none macOS or Linux, so there is no impossible-hardware pairing. On stock beta.30 those
presets silently reported 0, because the `Navigator.cpp` override is #697's unreleased work; the
touch signals were simply never emitted. On this branch they now produce a coherent touchscreen
laptop (`maxTouchPoints=10`, `TouchEvent` present, `(any-pointer: coarse)` true, `(pointer:
coarse)` false, `ontouchstart` absent). That is the fix doing its job, but it does mean roughly a
fifth of Windows preset draws now look different than they did.

It also closes a host leak in the other direction: on a machine with a real digitizer,
`dom.w3c_touch_events.enabled=2` autodetected true and exposed `TouchEvent` while the spoofed
`navigator.maxTouchPoints` said 0 — a direct contradiction sourced from the host. An explicit 0
now suppresses the interfaces to match.

The reference is the real capture from the XPS, read off the reference machine — not derived.
Worth noting what that corrected: an earlier reconstruction had the right *values* but the wrong
*set*. Fourteen of its sixteen matched the capture exactly; the other two were invented
(`'ontouchstart'` on `document` and `documentElement`, both false and harmless) and stood in for
two real signals that were missed — `document.createTouch` (false, gated by `LegacyAPIEnabled`
like `ontouchstart`) and `window.PointerEvent` (true, not gated on a digitizer at all). Both were
then measured on this build before being written down, so the match is on all sixteen recorded
signals rather than on fourteen guessed correctly.

`tests/assets/touch-reference.html` is the capture page, added so the recording can be
reproduced on another device.

## Verification

Built `152.0.4-beta.30` for `x86_64-pc-linux-gnu`, measured against a **stock beta.30 baseline**.

| suite | branch | stock | |
| --- | --- | --- | --- |
| Playwright (`tests/async/`) | 13 failed, 1083 passed | 14 failed, 1082 passed | no regressions |
| patch guards | 16 / 16 | — | 2 added |
| sundial (786 checks) | 6 fail | 6 fail | no regressions |
| `pythonlib/tests/` | 209 passed | — | |

The two new guards were both checked against a binary without their fix:
`touchscreen-digitizer.py` reports 12/16, and `search-service-init.py` fails all five of its
checks on stock beta.30. #755's guards were likewise run against stock:
`DEADLOCK: a mousemove at (31, 0) produced no hit-renderer ack` on stock, PASS here.

The three Playwright differences versus stock are all inside the known-flaky set documented in
`116a534` — `test_should_pause` appeared (it fails 5/5 on stock when run alone), and two others
passed this time.

Linux x86_64 only; sundial scanned on the `linux` profile only. Two notes for anyone
reproducing: sundial's failure *count* is not stable across runs — the two `CPU:` checks
redraw with `hardwareConcurrency`, giving 6 or 8 for the same build — so compare failure *sets*,
not totals. And under plain `headless=True` there is no WebGL context at all (true of stock
too), which fails both WebGL vectors on a build whose spoofing is working; `headless="virtual"`
with `CAMOUFOX_VIRTUAL_DISPLAY_SIZE` set gives a real context.

https://claude.ai/code/session_01W2RfR387Mh1JhZ9LZvkptP


JWriter20:
Human reviewed and tested against build on my fork. Will merge and create pre-release as there are some important fixes here.

PR #780, opened 2026-09-21
fix: derive navigator.buildID from spoofed UA version by default

dyiapanis:
## Problem

Camoufox spoofs `navigator.userAgent` to a chosen Firefox version, but `navigator.buildID` is only spoofed when set explicitly (`settings/properties.json` -> `camoucfg.jvv`). Unset, the engine falls back to XULAppInfo — the **host build's** own value. Distributed Camoufox builds ship with a stale one: our deployed v152.0.4-beta.30 instance reports `20181001000000` (a Firefox 63-era stamp) while the UA claims Firefox 152.

Any detector cross-checking claimed UA version against buildID recency sees a 2026 browser with a 2018 build. We hit this in production on Pixelscan-class checkers.

## Change

Derive a plausible `YYYYMMDDHHMMSS` from the spoofed major version when the user has not set one:

- Firefox 94 anchored at `2021-11-02`; each subsequent major +4 weeks (Mozilla cadence post-2021); fixed hour offset from the major so successive versions don't all stamp midnight
- Deterministic — the same identity always yields the same buildID
- Applied at both config-assembly points — `from_preset()` and `from_browserforge()` — covering preset, BrowserForge, and custom-fingerprint paths
- Explicit user-supplied values always win; a `buildID` carried inside a fingerprint preset is now honoured (previously dropped entirely by `from_preset()`)
- Versions before the anchor (Firefox < 94) are left alone rather than guessed

## Notes

- `camoucfg.jvv` validates `str[/^\d+$/]` — derived values conform
- No new dependencies; stdlib `datetime` only

## Testing

- New: `tests/test_preset_buildid.py` (10 tests) — derivation shape/determinism/monotonicity/anchor floor, both assembly paths, user-override precedence
- Full suite: **219 passed** (209 upstream + 10 new), 0 failures


dyiapanis:
Consolidation note: #735 (the earlier period-correct buildID PR, opt-in) is now closed as superseded by this one — the default-on UA-derived approach here is the stronger fix. All review attention can land on this PR.

